PR #17, opened 2020-03-09
Implement event stream backed by streaming gRPC service.

jmalloc:
@danilvpetrov This PR needs tests yet, but I thought I'd get you in particular to take a look because there's some very "fancy" context shuffling going on just to get the gRPC streaming interfaces to look like regular Go code and I know you've dealt with this before. 

I want to make sure the inline comments are really helpful in understanding it even if they are verbose (assuming the reader already understands `context.Context`).


jmalloc:
The stream interfaces in the `eventstream` package are very similar to those in [Aperture's `ordered` package](https://github.com/dogmatiq/aperture/blob/master/ordered/stream.go), but include full message meta-data, similar to Ax.

These interfaces are intended to provide an abstraction to event messages regardless of whether they come from the "local" persistence for "this" application, or across the network from another application.

api/stream.go
```diff
@@ -0,0 +1,222 @@
...
+
+// Close stops the cursor.
+//
+// Any current or future calls to Next() return a non-nil error.
+func (c *cursor) Close() error {
+	c.cancel()
+	return nil
+}
+
+// consume waits receives new messages, unmarshals them and pipes them over the
```
danilvpetrov:
```suggestion
// consume waits, receives new messages, unmarshals them, and pipes them over the
```

persistence/stream.go
```diff
@@ -0,0 +1,49 @@
...
+	// Any current or future calls to Next() return a non-nil error.
+	Close() error
+}
+
+// StreamMessage is a message on a stream.
+type StreamMessage struct {
+	// Offset is the offset of the message on the stream.
+	Offset uint64
+
+	// Envelope contains the message and its meta-data.
```
danilvpetrov:
```suggestion
	// Envelope contains the message and its metadata.
```

jmalloc: I've written this as `meta-data` everywhere else, so I'll leave that can of worms alone :)

api/stream.go
jmalloc: `waits` was an extra word... I've added your Oxford comma, though ;)

api/stream.go
```diff
@@ -0,0 +1,222 @@
...
+
+	// consumeCtx lives for the lifetime of the stream returned by the gRPC
+	// Consume() operation. This is how we cancel the gRPC stream, as it has no
+	// Close() method.
+	//
+	// It does NOT use ctx as a parent, because ctx is intended only to span the
+	// lifetime of this call to Open().
+	consumeCtx, cancelConsume := context.WithCancel(context.Background())
+
+	// done is signalling channel used to indicate that the call to Consume()
```
jmalloc:
```suggestion
	// done is a signalling channel used to indicate that the call to Consume()
```

jmalloc:
I will work this into a future PR.

PR #28, opened 2020-03-11
Prepare for SQL persistence.

jmalloc:
This PR brings in some helper functions and CI from projectionkit that is used to test with various SQL databases.


jmalloc:
Aside from the fact that the tests fail under CI for some reason, this implementation currently requires a DB connection per cursor, which is not really scalable.

It does this because it uses a temporary table to hold the potentially large number of message types in the cursor's filter.

I need to work out a better way of doing this, possibly by building a table of distinct query filters over time.

PR #44, opened 2020-03-15
Expose message types via messaging API.

jmalloc:
This PR exposes information about the message types accepted/produced by the `CommandQueue` and `EventStream` services in the `dogma.messaging.v1` API.

This information is more-specific that the information exposed via the `dogma.config.v1` API, because it includes the "portable" message names and MIME media-type information that is determined by the marshaler.

This could be used to do some HTTP-style content negotiation in the future, though I do not intend to implement anything like that in Infix.


api/messaging/eventstream/stream.go
```diff
@@ -23,24 +23,48 @@ import (
...
+	req := &messagingspec.MessageTypesRequest{
+		ApplicationKey: k,
+	}
+
+	res, err := es.MessageTypes(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	// Built a type-set containing any message types supported by the remote end
```
jmalloc:
```suggestion
	// Build a type-set containing any message types supported by the remote end
```

PR #119, opened 2020-04-10
Implement integration handlers.

jmalloc:
This PR adds support for Dogma integration handlers!

It also implements the functions mentioned in #135 and #136 and uses them in these new tests. I have not converted existing tests to use them yet.


jmalloc:
Heh, without tests this PR actually reduces LOC.

PR #148, opened 2020-04-13
Add offsetstore persistence subsystem.

danilvpetrov:
This PR adds offset persistence subsystem. The offset tracking is designed to track the offset of the application that the current application is interacting with.

### DIFF
--- a/persistence/datastore.go
+++ b/persistence/datastore.go
@@ -28,7 +28,7 @@ type DataStore interface {
 	// QueueStoreRepository returns the application's queue store repository.
 	QueueStoreRepository() queuestore.Repository
 
-	// OffsetStoreRepository returns the application's offset repository.
+	// OffsetStoreRepository returns the application's offset store repository.
 	OffsetStoreRepository() offsetstore.Repository
 
 	// Begin starts a new transaction.
--- a/persistence/datastore.go
+++ b/persistence/datastore.go
@@ -25,11 +25,11 @@ type DataStore interface {
 	// EventStoreRepository returns the application's event store repository.
 	EventStoreRepository() eventstore.Repository
 
-	// QueueStoreRepository returns the application's queue store repository.
-	QueueStoreRepository() queuestore.Repository
-
 	// OffsetStoreRepository returns the application's offset store repository.
 	OffsetStoreRepository() offsetstore.Repository
+
+	// QueueStoreRepository returns the application's queue store repository.
+	QueueStoreRepository() queuestore.Repository
 
 	// Begin starts a new transaction.
 	Begin(ctx context.Context) (Transaction, error)
--- a/persistence/datastore.go
+++ b/persistence/datastore.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/dogmatiq/infix/persistence/subsystem/aggregatestore"
 	"github.com/dogmatiq/infix/persistence/subsystem/eventstore"
+	"github.com/dogmatiq/infix/persistence/subsystem/offsetstore"
 	"github.com/dogmatiq/infix/persistence/subsystem/queuestore"
 	"go.uber.org/multierr"
 )
@@ -26,6 +27,9 @@ type DataStore interface {
 
 	// QueueStoreRepository returns the application's queue store repository.
 	QueueStoreRepository() queuestore.Repository
+
+	// OffsetStoreRepository returns the application's offset repository.
+	OffsetStoreRepository() offsetstore.Repository
 
 	// Begin starts a new transaction.
 	Begin(ctx context.Context) (Transaction, error)
--- a/persistence/internal/providertest/offsetstore/repository.go
+++ b/persistence/internal/providertest/offsetstore/repository.go
@@ -1,0 +1,41 @@
+package offsetstore
+
+import (
+	"context"
+
+	"github.com/dogmatiq/infix/persistence"
+	"github.com/dogmatiq/infix/persistence/internal/providertest/common"
+	"github.com/dogmatiq/infix/persistence/subsystem/offsetstore"
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+)
+
+// DeclareRepositoryTests declares a functional test-suite for a specific
+// offsetstore.Repository implementation.
+func DeclareRepositoryTests(tc *common.TestContext) {
+	ginkgo.Describe("type offsetstore.Repository", func() {
+		var (
+			dataStore  persistence.DataStore
+			repository offsetstore.Repository
+			tearDown   func()
+		)
+
+		ginkgo.BeforeEach(func() {
+			dataStore, tearDown = tc.SetupDataStore()
+			repository = dataStore.OffsetStoreRepository()
+
+		})
+
+		ginkgo.AfterEach(func() {
+			tearDown()
+		})
+
+		ginkgo.Describe("func LoadOffset()", func() {
+			ginkgo.It("loads the initial offset as zero", func() {
+				o, err := repository.LoadOffset(context.Background(), "<app-key>")
+				gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+				gomega.Expect(o).Should(gomega.BeNumerically("==", 0))
+			})
+		})
+	})
+}
--- a/persistence/internal/providertest/offsetstore/repository.go
+++ b/persistence/internal/providertest/offsetstore/repository.go
@@ -31,7 +31,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 
 		ginkgo.Describe("func LoadOffset()", func() {
 			ginkgo.It("loads the initial offset as zero", func() {
-				o, err := repository.LoadOffset(context.Background(), "<app-key>")
+				o, err := repository.LoadOffset(context.Background(), "<source-app-key>")
 				gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
 				gomega.Expect(o).Should(gomega.BeNumerically("==", 0))
 			})
--- a/persistence/internal/providertest/offsetstore/repository.go
+++ b/persistence/internal/providertest/offsetstore/repository.go
@@ -33,7 +33,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 			ginkgo.When("application has no previous offsets associated", func() {
 				ginkgo.It("loads the initial offset as zero", func() {
 					actual, err := repository.LoadOffset(
-						context.Background(),
+						tc.Context,
 						"<source-app-key>",
 					)
 					gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
@@ -48,7 +48,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 					saveOffset(tc.Context, dataStore, "<source-app-key>", c, n)
 
 					actual, err := repository.LoadOffset(
-						context.Background(),
+						tc.Context,
 						"<source-app-key>",
 					)
 					gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
--- a/persistence/internal/providertest/offsetstore/repository.go
+++ b/persistence/internal/providertest/offsetstore/repository.go
@@ -30,21 +30,40 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 		})
 
 		ginkgo.Describe("func LoadOffset()", func() {
-			ginkgo.It("loads the initial offset as zero", func() {
-				o, err := repository.LoadOffset(context.Background(), "<source-app-key>")
-				gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-				gomega.Expect(o).Should(gomega.BeNumerically("==", 0))
+			ginkgo.When("application has no previous offsets associated", func() {
+				ginkgo.It("loads the initial offset as zero", func() {
+					actual, err := repository.LoadOffset(
+						context.Background(),
+						"<source-app-key>",
+					)
+					gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+					gomega.Expect(actual).Should(gomega.BeNumerically("==", 0))
+				})
 			})
 
-			// TO-DO: add a test with transaction involved to test the
-			// incremented offset.
+			ginkgo.When("application has previous offsets associated", func() {
+				ginkgo.It("returns the current offset", func() {
+					c := offsetstore.Offset(0)
+					n := offsetstore.Offset(1)
+					saveOffset(tc.Context, dataStore, "<source-app-key>", c, n)
 
-			ginkgo.It("returns an error if the context is canceled", func() {
-				ctx, cancel := context.WithCancel(tc.Context)
-				cancel()
+					actual, err := repository.LoadOffset(
+						context.Background(),
+						"<source-app-key>",
+					)
+					gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+					gomega.Expect(actual).Should(gomega.BeNumerically("==", n))
+				})
+			})
 
-				_, err := repository.LoadOffset(ctx, "<app-key>")
-				gomega.Expect(err).To(gomega.Equal(context.Canceled))
+			ginkgo.When("context is cancelled", func() {
+				ginkgo.It("returns the context cancellation error", func() {
+					ctx, cancel := context.WithCancel(tc.Context)
+					cancel()
+
+					_, err := repository.LoadOffset(ctx, "<source-app-key>")
+					gomega.Expect(err).To(gomega.Equal(context.Canceled))
+				})
 			})
 		})
 	})
--- a/persistence/internal/providertest/offsetstore/repository.go
+++ b/persistence/internal/providertest/offsetstore/repository.go
@@ -37,7 +37,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 						"<source-app-key>",
 					)
 					gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-					gomega.Expect(actual).Should(gomega.BeNumerically("==", 0))
+					gomega.Expect(actual).To(gomega.BeEquivalentTo(0))
 				})
 			})
 
@@ -52,7 +52,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 						"<source-app-key>",
 					)
 					gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-					gomega.Expect(actual).Should(gomega.BeNumerically("==", n))
+					gomega.Expect(actual).To(gomega.BeEquivalentTo(n))
 				})
 			})
 
--- a/persistence/internal/providertest/offsetstore/repository.go
+++ b/persistence/internal/providertest/offsetstore/repository.go
@@ -3,6 +3,7 @@ package offsetstore
 import (
 	"context"
 
+	"github.com/dogmatiq/infix/eventstream"
 	"github.com/dogmatiq/infix/persistence"
 	"github.com/dogmatiq/infix/persistence/internal/providertest/common"
 	"github.com/dogmatiq/infix/persistence/subsystem/offsetstore"
@@ -43,8 +44,8 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 
 			ginkgo.When("application has previous offsets associated", func() {
 				ginkgo.It("returns the current offset", func() {
-					c := offsetstore.Offset(0)
-					n := offsetstore.Offset(1)
+					c := eventstream.Offset(0)
+					n := eventstream.Offset(1)
 					saveOffset(tc.Context, dataStore, "<source-app-key>", c, n)
 
 					actual, err := repository.LoadOffset(
--- a/persistence/internal/providertest/offsetstore/repository.go
+++ b/persistence/internal/providertest/offsetstore/repository.go
@@ -23,7 +23,6 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 		ginkgo.BeforeEach(func() {
 			dataStore, tearDown = tc.SetupDataStore()
 			repository = dataStore.OffsetStoreRepository()
-
 		})
 
 		ginkgo.AfterEach(func() {
@@ -35,6 +34,17 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 				o, err := repository.LoadOffset(context.Background(), "<app-key>")
 				gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
 				gomega.Expect(o).Should(gomega.BeNumerically("==", 0))
+			})
+
+			// TO-DO: add a test with transaction involved to test the
+			// incremented offset.
+
+			ginkgo.It("returns an error if the context is canceled", func() {
+				ctx, cancel := context.WithCancel(tc.Context)
+				cancel()
+
+				_, err := repository.LoadOffset(ctx, "<app-key>")
+				gomega.Expect(err).To(gomega.Equal(context.Canceled))
 			})
 		})
 	})
--- a/persistence/internal/providertest/offsetstore/transaction.go
+++ b/persistence/internal/providertest/offsetstore/transaction.go
@@ -3,6 +3,7 @@ package offsetstore
 import (
 	"sync"
 
+	"github.com/dogmatiq/infix/eventstream"
 	"github.com/dogmatiq/infix/persistence"
 	"github.com/dogmatiq/infix/persistence/internal/providertest/common"
 	"github.com/dogmatiq/infix/persistence/subsystem/offsetstore"
@@ -33,8 +34,8 @@ func DeclareTransactionTests(tc *common.TestContext) {
 		ginkgo.Describe("func SaveOffset()", func() {
 			ginkgo.When("application has no previous offsets associated", func() {
 				ginkgo.It("saves an offset", func() {
-					c := offsetstore.Offset(0)
-					n := offsetstore.Offset(1)
+					c := eventstream.Offset(0)
+					n := eventstream.Offset(1)
 					saveOffset(tc.Context, dataStore, "<source-app-key>", c, n)
 
 					actual := loadOffset(tc.Context, repository, "<source-app-key>")
@@ -42,8 +43,8 @@ func DeclareTransactionTests(tc *common.TestContext) {
 				})
 
 				ginkgo.It("it does not update the offset when an OCC conflict occurs", func() {
-					c := offsetstore.Offset(1)
-					n := offsetstore.Offset(2)
+					c := eventstream.Offset(1)
+					n := eventstream.Offset(2)
 
 					err := persistence.WithTransaction(
 						tc.Context,
@@ -59,8 +60,8 @@ func DeclareTransactionTests(tc *common.TestContext) {
 				})
 
 				ginkgo.It("does not save the offset when the transaction is rolled-back", func() {
-					c := offsetstore.Offset(0)
-					n := offsetstore.Offset(1)
+					c := eventstream.Offset(0)
+					n := eventstream.Offset(1)
 
 					err := common.WithTransactionRollback(
 						tc.Context,
@@ -83,11 +84,11 @@ func DeclareTransactionTests(tc *common.TestContext) {
 
 			ginkgo.When("application has previous offsets associated", func() {
 				var (
-					current offsetstore.Offset
+					current eventstream.Offset
 				)
 
 				ginkgo.BeforeEach(func() {
-					current = offsetstore.Offset(0)
+					current = eventstream.Offset(0)
 					saveOffset(tc.Context, dataStore, "<source-app-key>", current, 5)
 					current = 5
 				})
@@ -101,7 +102,7 @@ func DeclareTransactionTests(tc *common.TestContext) {
 
 				table.DescribeTable(
 					"it does not update the offset when an OCC conflict occurs",
-					func(conflictingOffset offsetstore.Offset) {
+					func(conflictingOffset eventstream.Offset) {
 						err := persistence.WithTransaction(
 							tc.Context,
 							dataStore,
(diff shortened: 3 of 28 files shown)


persistence/internal/providertest/offsetstore/repository.go
```diff
@@ -0,0 +1,41 @@
...
+		var (
+			dataStore  persistence.DataStore
+			repository offsetstore.Repository
+			tearDown   func()
+		)
+
+		ginkgo.BeforeEach(func() {
+			dataStore, tearDown = tc.SetupDataStore()
+			repository = dataStore.OffsetStoreRepository()
+
```
jmalloc, line 26:
```suggestion

```

persistence/internal/providertest/offsetstore/repository.go
```diff
@@ -0,0 +1,41 @@
...
+		})
+
+		ginkgo.AfterEach(func() {
+			tearDown()
+		})
+
+		ginkgo.Describe("func LoadOffset()", func() {
+			ginkgo.It("loads the initial offset as zero", func() {
+				o, err := repository.LoadOffset(context.Background(), "<app-key>")
```
jmalloc, line 35: I would suggest using a different string here, because all of the data-store that gets setup is for the app known as `<app-key>`. This way we make sure we're not swapping these values inadvertently.

The "source application" is existing terminology I have meaning the application that produced a message. In this case we're getting messages from the stream of their source application.

```suggestion
				o, err := repository.LoadOffset(context.Background(), "<source-app-key>")
```

persistence/internal/providertest/offsetstore/transaction.go
```diff
@@ -0,0 +1,63 @@
...
+		var (
+			dataStore  persistence.DataStore
+			repository offsetstore.Repository
+			tearDown   func()
+		)
+
+		ginkgo.BeforeEach(func() {
+			dataStore, tearDown = tc.SetupDataStore()
+			repository = dataStore.OffsetStoreRepository()
+
```
jmalloc, line 24:
```suggestion

```

persistence/internal/providertest/offsetstore/transaction.go
```diff
@@ -0,0 +1,63 @@
...
+		})
+
+		ginkgo.AfterEach(func() {
+			tearDown()
+		})
+
+		ginkgo.Describe("func SaveOffset()", func() {
+			ginkgo.It("safes an offset", func() {
```
jmalloc, line 32:
```suggestion
			ginkgo.It("saves an offset", func() {
```

persistence/datastore.go
```diff
@@ -23,6 +24,9 @@ type DataStore interface {
 	// QueueStoreRepository returns the application's queue store repository.
 	QueueStoreRepository() queuestore.Repository
 
+	// OffsetStoreRepository returns the application's offset repository.
```
jmalloc, line 27:
```suggestion
	// OffsetStoreRepository returns the application's offset store repository.
```

persistence/subsystem/offsetstore/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package offsetstore defines an API for persisting the application event
+// stream offset.
```
jmalloc, line 2:
```suggestion
// Package offsetstore defines an API for persisting event offsets
// representing "progress" through an application's event stream.
```

эти трудно хорошо объяснить, может быть этот текст лучше... я не уверен

persistence/subsystem/offsetstore/offset.go
```diff
@@ -0,0 +1,4 @@
+package offsetstore
+
+// Offset is the position of application's event stream offset within the store.
```
jmalloc, line 3:
```suggestion
// Offset is the position of an event within an application's event stream.
```

persistence/subsystem/offsetstore/repository.go
```diff
@@ -0,0 +1,13 @@
+package offsetstore
+
+import (
+	"context"
+)
+
+// Repository is an interface for reading a persisted event stream offset of the
+// application in question.
```
jmalloc, line 8:
```suggestion
// Repository is an interface for reading persisted offsets.
```

persistence/subsystem/offsetstore/repository.go
```diff
@@ -0,0 +1,13 @@
...
+type Repository interface {
+	// LoadOffset loads the offset of the application with the key ak from the
+	// repository.
```
jmalloc, line 11:
```suggestion
	// LoadOffset loads the offset associated with a specific application.
	//
	// ak is the application's identity key.
```

persistence/subsystem/offsetstore/transaction.go
```diff
@@ -0,0 +1,25 @@
+package offsetstore
+
+import (
+	"context"
+	"errors"
+)
+
+// ErrConflict is returned when the offset supplied as a parameter to method
+// Transaction.SaveOffset does not equal to the currently persisted offset.
+var ErrConflict = errors.New("optimistic concurrency conflict in the application event stream offset")
```
jmalloc, line 10:
```suggestion
var ErrConflict = errors.New("an optimistic concurrency conflict occured while persisting to the offset store")
```

I will change `ErrConflict` in `queuestore` to match this kind of phrasing.

persistence/subsystem/offsetstore/transaction.go
```diff
@@ -0,0 +1,25 @@
+package offsetstore
+
+import (
+	"context"
+	"errors"
+)
+
+// ErrConflict is returned when the offset supplied as a parameter to method
+// Transaction.SaveOffset does not equal to the currently persisted offset.
```
jmalloc, line 9:
```suggestion
// ErrConflict is returned by transaction operations when a persisted
// offset can not be updated because the supplied "current" offset is
// out of date.
```

persistence/subsystem/offsetstore/transaction.go
```diff
@@ -0,0 +1,25 @@
...
+type Transaction interface {
+	// SaveOffset persists the application event stream offset.
+	//
+	// If the given offset o does not equal to the currently persisted offset,
+	// an optimistic concurrency conflict has occurred and ErrConflict is
+	// returned.
+	SaveOffset(
+		ctx context.Context,
+		ak string,
+		o Offset,
```
jmalloc, line 23:
```suggestion
		c, n Offset,
```

We still need the caller to supply both the "current" and "next" offsets. These variable names match conceptually similar ideas from the projection interface, etc.

persistence/subsystem/offsetstore/transaction.go
```diff
@@ -0,0 +1,25 @@
...
+var ErrConflict = errors.New("optimistic concurrency conflict in the application event stream offset")
+
+// Transaction defines the primitive persistence operations for manipulating
+// the application event stream offset.
+type Transaction interface {
+	// SaveOffset persists the application event stream offset.
+	//
+	// If the given offset o does not equal to the currently persisted offset,
+	// an optimistic concurrency conflict has occurred and ErrConflict is
+	// returned.
```
jmalloc, line 19:
```suggestion
	// SaveOffset persists the "next" offset to be consumed for a specific
	// application.
	//
	// ak is the application's identity key.
	//
	// c must be the offset currently associated with ak, otherwise an optimistic
	// concurrency conflict has occurred, the offset is not saved and ErrConflict
	// is returned.
```

jmalloc:
I linked to some tests from `queuestore` that should have equivalents in this PR, in no particular order. Notice though that I make use of `When()` blocks in the other tests for scenarios like "when the message is already in the queue"... you could have similar ones here like "when the application already has an associated offset", etc.

These combinations blow out the tests a bit, but it's important that we actually verify "inserts and updates" both work, in different orders, etc even though it is obvious from the SQL implementations, it's not as obvious for BoltDB and the memory ones.

danilvpetrov:
OK, thanks I will get down to those tests.

persistence/internal/providertest/offsetstore/transaction.go
```diff
@@ -0,0 +1,198 @@
...
+					gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+					gomega.Expect(actual).Should(gomega.BeNumerically("==", n))
+				})
+			})
+
+			ginkgo.When("an OCC conflict occurs", func() {
+				table.DescribeTable(
+					"it does not update the offset",
+					func(conflictingCurrentOffset, nextOffset offsetstore.Offset) {
+						// Update the current offset twice so that it's up to 2.
```
jmalloc, line 109: You don't really need to do this double-update here, it was only necessary in the `queuestore` test because each update incremented the revision by `1`.

persistence/internal/providertest/offsetstore/transaction.go
```diff
@@ -0,0 +1,198 @@
...
+						},
+					)
+
+					actual, err := repository.LoadOffset(tc.Context, "<source-app-key>")
+					gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+					gomega.Expect(actual).Should(gomega.BeNumerically("==", n))
+				})
+			})
+
+			ginkgo.When("an OCC conflict occurs", func() {
```
jmalloc, line 105: Some kind of OCC testing also needs to be performed in the `when application has no previous offsets associated` block. I'd retain this table test but move it to the`when application has previous offsets associated` section.

persistence/internal/providertest/offsetstore/transaction.go
```diff
@@ -0,0 +1,198 @@
...
+			repository = dataStore.OffsetStoreRepository()
+		})
+
+		ginkgo.AfterEach(func() {
+			tearDown()
+		})
+
+		ginkgo.Describe("func SaveOffset()", func() {
+			ginkgo.When("application has no previous offsets associated", func() {
+				ginkgo.It("saves an offset with the current offset equal to zero", func() {
```
jmalloc, line 35:
```suggestion
				ginkgo.It("saves the offset", func() {
```

This test doesn't need a lengthy description, it's just the common case, and will make sense when you see it next to the OCC test (see comment below).

persistence/internal/providertest/offsetstore/transaction.go
```diff
@@ -0,0 +1,198 @@
...
+					saveOffset(tc.Context, dataStore, "<source-app-key>", c, n)
+
+					actual, err := repository.LoadOffset(tc.Context, "<source-app-key>")
+					gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+					gomega.Expect(actual).Should(gomega.BeNumerically("==", n))
+				})
+			})
+
+			ginkgo.When("application has previous offsets associated", func() {
+				ginkgo.It("saves an offset with the current offset passed", func() {
```
jmalloc, line 47:
```suggestion
				ginkgo.It("updates the offset", func() {
```

persistence/internal/providertest/offsetstore/transaction.go
```diff
@@ -0,0 +1,198 @@
...
+					n += n
+
+					saveOffset(tc.Context, dataStore, "<source-app-key>", c, n)
+
+					actual, err := repository.LoadOffset(tc.Context, "<source-app-key>")
+					gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+					gomega.Expect(actual).Should(gomega.BeNumerically("==", n))
+				})
+
+				ginkgo.It("does not update the current offset with the equal next offset", func() {
```
jmalloc, line 62: I'm not sure about this one. What is it guaranteeing? It's not so much that it "does not update", perhaps, but that there's no error if `c == n`?

If I have that right, it can probably be removed as "not important to the requirements of the interface", but it's not harmful either.

persistence/internal/providertest/offsetstore/transaction.go
```diff
@@ -0,0 +1,198 @@
...
+							dataStore,
+							func(tx persistence.ManagedTransaction) error {
+								return tx.SaveOffset(tc.Context, ak, c, n)
+							},
+						)
+						gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+					}
+
+					g.Add(3)
+					go fn("<source-app1-key>", 0, 1)
```
jmalloc, line 190: Maybe use distinct "next" values for each app here, so we are sure they don't get swapped somehow.

persistence/internal/providertest/offsetstore/transaction.go
```diff
@@ -32,88 +32,68 @@ func DeclareTransactionTests(tc *common.TestContext) {
...
-			})
 
-			ginkgo.When("an OCC conflict occurs", func() {
 				table.DescribeTable(
-					"it does not update the offset",
+					"it does not update the offset when an OCC conflict occurs",
 					func(conflictingCurrentOffset, nextOffset offsetstore.Offset) {
-						// Update the current offset twice so that it's up to 2.
-						// Otherwise we can't test for 1 as a too-low value.
+						// Update the current offset twice so that it's up to 1.
```
jmalloc, line 93:
```suggestion
```

It's not being updated twice ;)

persistence/internal/providertest/offsetstore/transaction.go
```diff
@@ -0,0 +1,205 @@
...
+							},
+						)
+						gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+					}
+
+					g.Add(3)
+					go fn("<source-app1-key>", 0, 1)
+					go fn("<source-app2-key>", 0, 2)
+					go fn("<source-app3-key>", 0, 3)
+					g.Wait()
```
jmalloc, line 200: This test needs to assert that the expected offsets were actually stored against the right keys and all that, after the goroutines are finished.

persistence/internal/providertest/offsetstore/repository.go
```diff
@@ -0,0 +1,70 @@
...
+
+		ginkgo.AfterEach(func() {
+			tearDown()
+		})
+
+		ginkgo.Describe("func LoadOffset()", func() {
+			ginkgo.When("application has no previous offsets associated", func() {
+				ginkgo.It("loads the initial offset as zero", func() {
+					actual, err := repository.LoadOffset(
+						context.Background(),
```
jmalloc, line 36:
```suggestion
						tc.Context,
```

The test context has a `context.Context` in it that handles timing out the tests, etc.

jmalloc: Oh you've used it elsewhere, just this one.

persistence/internal/providertest/offsetstore/transaction.go
```diff
@@ -0,0 +1,205 @@
...
+							gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+
+							return nil
+						},
+					)
+
+					actual, err := repository.LoadOffset(tc.Context, "<source-app-key>")
+					gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+					gomega.Expect(actual).To(gomega.BeEquivalentTo(n))
+				})
```
jmalloc, line 150: There's another test in the queuestore called `uses the uncommitted revision for OCC checks` which needs an analogue here.

persistence/internal/providertest/offsetstore/transaction.go
jmalloc: Oh, I guess it is! sorry man, I think you can get away with a `BeforeEach()` here though that does the initial update from `0` for both this table test and the `updates the offset` test.

persistence/internal/providertest/offsetstore/transaction.go
```diff
@@ -0,0 +1,205 @@
...
+			})
+
+			ginkgo.When("the transaction is rolled-back", func() {
+				ginkgo.It("does not save the offset", func() {
```
jmalloc, line 154: Man, I apologize, I know this is tedious, but this test also needs to appear in both the `application has no previous offsets associated` and `application has previous offsets associated` blocks.

persistence/internal/providertest/offsetstore/transaction.go
```diff
@@ -0,0 +1,205 @@
...
+							c = n
+							n += 123
+
+							err = tx.SaveOffset(tc.Context, "<source-app-key>", c, n)
+							gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+
+							return nil
+						},
+					)
+
```
jmalloc, line 146:
```suggestion
gomega.Expect(err).ShouldNot(gomega.HaveOccurred())

```

persistence/internal/providertest/offsetstore/transaction.go
danilvpetrov: Hey man, just to specify, which `When` block this should go? I assume it should go to `ginkgo.When("called multiple times within the same transaction"` similar to `queuestore` tests?

persistence/internal/providertest/offsetstore/transaction.go
danilvpetrov: So the both table tests (one in `application has no previous offsets associated` section, another in `application has previous offsets associated`) should not internal increment of the offset, right? But the one in `application has previous offsets associated` will rely on the incremented offset in `BeforeEach()`, correct? What about table entries should they remain as they are for both table tests?

jmalloc: Regarding the offsets and `BeforeEach` yep that's what I meant, that should be neater.

I don't think the test in `application has no previous offsets associated` needs to be table-driven at all because there's no special case - any incorrect offset you give it is non-zero and too high. That's how the tests in #163 and the `queuestore` are structured.

For the `application has previous offsets associated` tests, you can bump "up past 1" in the single save done in `BeforeEach()`. Then I think you just need the `zero`, `too low` and `too high` test cases. There's probably no need to pass in the "new" offset in the table entries, it can just be hardcoded to some arbitrary high number (good ol' `123`) in the test function.

persistence/internal/providertest/offsetstore/transaction.go
jmalloc: Yep, that's right - it's checking that the OCC is based on the previous (as yet uncommitted) change in the same tx.

persistence/subsystem/offsetstore/offset.go
```diff
@@ -0,0 +1,4 @@
+package offsetstore
+
+// Offset is the position of an event within an application's event stream.
+type Offset uint64
```
jmalloc, line 4:
```suggestion
```

As per #172 this type should be removed and you should instead reference `eventstream.Offset` directly. I haven't actually *done* anything with #172 yet it's just highlighting a package structure that seems odd.

danilvpetrov: No worries, I can do that in my next commit.

persistence/provider/memory/offsetstore.go
```diff
@@ -0,0 +1,102 @@
...
+
+// stageSave adds a "SaveOffset" operation to the change-set.
+//
+// It returns false if there is an OCC conflict.
+func (cs *offsetStoreChangeSet) stageSave(
+	db *offsetStoreDatabase,
+	ak string,
+	c, n eventstream.Offset,
+) bool {
+
```
koden-km, line 42: Empty line.
```suggestion
```

persistence/provider/sql/postgres/error.go
```diff
@@ -170,6 +171,56 @@ func (d errorConverter) ScanEvent(
 	return d.d.ScanEvent(rows, i)
 }
 
+//
```
koden-km, line 174: Shouldn't these doc comments be on the functions?

jmalloc: The

```
//
// offsetstore
//
``` 

is just a separator between methods from different interfaces, but the `// LoadOffset` docs can be omitted entirely from this file.

persistence/provider/sql/postgres/error.go
```diff
@@ -170,6 +171,56 @@ func (d errorConverter) ScanEvent(
...
+// InsertOffset inserts a new offset associated with the given application
+// key.
+//
+// It returns false if the row already exists.
+
+// UpdateOffset updates the offset associated with the given application
+// key.
+//
+// It returns false if the row does not exist or c is not the current
+// offset associated with the given application key.
```
jmalloc, line 193:
```suggestion
```

persistence/provider/sql/postgres/error.go
jmalloc: For clarity: I leave the docs out of this file because it's just another place to keep them in sync, and this file doesn't actually implement any of the logic described in the docs, it's just a passthrough that fixes some bung errors that the postgresql driver returns.

persistence/provider/boltdb/datastore.go
```diff
@@ -42,6 +43,11 @@ func (ds *dataStore) EventStoreRepository() eventstore.Repository {
 	return &eventStoreRepository{ds.db, ds.appKey}
 }
 
+// OffsetStoreRepository returns the application's event store repository.
+func (ds *dataStore) OffsetStoreRepository() offsetstore.Repository {
+	return &offsetStoreRepository{ds.db}
```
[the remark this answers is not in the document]
danilvpetrov: The [new commit](https://github.com/dogmatiq/infix/commit/bc6f7ec8f3d67cf0fbbdbc2fc78b17f939ad2f0f) brings the 'owner' application key to all 'offsetstore' implementations (except in-memory package) as we discussed over Slack.

persistence/provider/memory/offsetstore.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil
+	}
+
+	return offsetstore.ErrConflict
+}
+
+// offsetStoreChangeSet contains modifications to the offset store that have
+// been performed within a transaction but not yet committed.
+type offsetStoreChangeSet struct {
+	items map[string]eventstream.Offset
```
jmalloc, line 31:
```suggestion
	offsets map[string]eventstream.Offset
```

persistence/provider/memory/offsetstore.go
```diff
@@ -0,0 +1,101 @@
...
+// stageSave adds a "SaveOffset" operation to the change-set.
+//
+// It returns false if there is an OCC conflict.
+func (cs *offsetStoreChangeSet) stageSave(
+	db *offsetStoreDatabase,
+	ak string,
+	c, n eventstream.Offset,
+) bool {
+	// Get both the committed item, and the item as it appears with its changes
+	// staged in this change-set.
```
jmalloc, line 43:
```suggestion
	// Get both the committed offset, and the staged offset from this change-set.
```

Yeah I know, hyper pedantic again, but there's nothing called an "item" here, blah blah :)

persistence/provider/memory/offsetstore.go
```diff
@@ -0,0 +1,101 @@
...
+	committed := db.items[ak]
+	staged, changed := cs.items[ak]
+
+	// The "effective" item is how the item appears to this transaction.
```
jmalloc, line 47:
```suggestion
	// The "effective" value is how the offset appears to this transaction.
```

persistence/provider/memory/offsetstore.go
```diff
@@ -0,0 +1,101 @@
...
+		return 0, err
+	}
+	defer r.db.RUnlock()
+
+	return r.db.offset.items[ak], nil
+}
+
+// offsetStoreDatabase contains data that is committed to the offset store.
+type offsetStoreDatabase struct {
+	items map[string]eventstream.Offset
```
jmalloc, line 89:
```suggestion
	offsets map[string]eventstream.Offset
```

persistence/provider/sql/sqlite/offsetstore.go
```diff
@@ -0,0 +1,86 @@
...
+func (driver) LoadOffset(
+	ctx context.Context,
+	db *sql.DB,
+	ak, sk string,
+) (eventstream.Offset, error) {
+	var o eventstream.Offset
+
+	row := db.QueryRowContext(
+		ctx,
+		`SELECT next_offset
```
jmalloc, line 25:
```suggestion
		`SELECT
			next_offset
```

persistence/provider/sql/sqlite/offsetstore.go
```diff
@@ -0,0 +1,86 @@
...
+		FROM offset_store
+		WHERE app_key = $1 AND source_app_key = $2`,
+		ak, sk,
```
jmalloc, line 28:
```suggestion
		WHERE app_key = $1
		AND source_app_key = $2`,
		ak,
		sk,
```

persistence/provider/sql/offsetstore.go
```diff
@@ -0,0 +1,99 @@
...
+
+// LoadOffset loads the offset associated with a specific application.
+func (r *offsetStoreRepository) LoadOffset(
+	ctx context.Context,
+	ak string,
+) (eventstream.Offset, error) {
+	o, err := r.d.LoadOffset(ctx, r.db, r.appKey, ak)
+	if err == sql.ErrNoRows {
+		return 0, nil
+	}
```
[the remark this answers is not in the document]
danilvpetrov: Ok, hopefully it's fixed in the [last commit](https://github.com/dogmatiq/infix/pull/148/commits/328340be952c0dd685fe28cd4448d7f62f622578).

persistence/provider/sql/mysql/offsetstore.go
```diff
@@ -0,0 +1,81 @@
...
+	ak, sk string,
+) (eventstream.Offset, error) {
+	var o eventstream.Offset
+
+	row := db.QueryRowContext(
+		ctx,
+		`SELECT next_offset
+		FROM offset_store
+		WHERE app_key = ? AND source_app_key = ?`,
+		ak, sk,
```
jmalloc, line 28: 🐧 **This suggestion brought to you by pedantic penguin.**

```suggestion
		`SELECT
			next_offset
		FROM offset_store
		WHERE app_key = ?
		AND source_app_key = ?`,
		ak,
		sk,
```

persistence/provider/sql/postgres/offsetstore.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if err := row.Scan(&o); err != nil {
+		if err == sql.ErrNoRows {
+			return 0, nil
+		}
+
+		return 0, err
+	}
+
+	return o, nil
```
jmalloc, line 39:
```suggestion
	var o eventstream.Offset
	err := row.Scan(&o)
	if err == sql.ErrNoRows {
		err = nil
	}

	return o, err
```

persistence/provider/sql/postgres/offsetstore.go
```diff
@@ -0,0 +1,92 @@
...
+//
+// If there is no offset associated with the given source application key,
+// the offset is returned as zero and error as nil.
+func (driver) LoadOffset(
+	ctx context.Context,
+	db *sql.DB,
+	ak, sk string,
+) (eventstream.Offset, error) {
+	var o eventstream.Offset
+
```
jmalloc, line 22:
```suggestion
```

persistence/provider/sql/mysql/offsetstore.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if err := row.Scan(&o); err != nil {
+		if err == sql.ErrNoRows {
+			return 0, nil
+		}
+
+		return 0, err
+	}
+
+	return o, nil
```
jmalloc, line 39: This is the other half of the suggestion at the top of this function that move the declaration of `o`. I'm also just trying to eliminate unnecessary branching, however minor. IMO this makes it clearer that in the context of this operation `ErrNoRows` is not an error.

```suggestion
	var o eventstream.Offset
	err := row.Scan(&o)
	if err == sql.ErrNoRows {
		err = nil
	}

	return o, err
```

persistence/provider/sql/mysql/offsetstore.go
```diff
@@ -0,0 +1,81 @@
...
+//
+// If there is no offset associated with the given source application key,
+// the offset is returned as zero and error as nil.
+func (driver) LoadOffset(
+	ctx context.Context,
+	db *sql.DB,
+	ak, sk string,
+) (eventstream.Offset, error) {
+	var o eventstream.Offset
+
```
jmalloc, line 22: I've moved this variable declaration down to the scan. I'm invoking the "declare variables close to their usage" adage, here. I've made the same suggestion in the other SQL packages too.

```suggestion
```

persistence/provider/sql/mysql/offsetstore.go
```diff
@@ -0,0 +1,81 @@
...
+	ctx context.Context,
+	tx *sql.Tx,
+	ak, sk string,
+	n eventstream.Offset,
+) (bool, error) {
+	return insertIgnore(
+		ctx,
+		tx,
+		`INSERT INTO offset_store VALUES(?, ?, ?)`,
+		ak, sk, n,
```
jmalloc, line 56: Please **always** use explicit column names in queries. Even though this is covered by tests within this repo, some way down the line someone will modify some production database as part of some weird hacky migration and individual queries will fail in weird and wonderful ways, especially with MySQL which seems to like to guess what you mean half the time :)

MySQL also has this "update-like" insert syntax, which I think is actually far more pleasant to work with than the `VALUES` tuple.

```suggestion
		`INSERT INTO offset_store SET
			app_key_id = ?,
			source_app_key = ?,
			next_offset = ?`,
		ak,
		sk,
		n,
```

persistence/provider/sql/mysql/offsetstore.go
```diff
@@ -0,0 +1,81 @@
...
+) (_ bool, err error) {
+	defer sqlx.Recover(&err)
+
+	return sqlx.TryExecRow(
+		ctx,
+		tx,
+		`UPDATE offset_store
+		SET next_offset = ?
+		WHERE app_key = ? AND source_app_key = ? AND next_offset = ?`,
+		n, ak, sk, c,
```
jmalloc, line 79: 🐧 **This suggestion brought to you by pedantic penguin.**

```suggestion
		`UPDATE offset_store SET
			next_offset = ?
		WHERE app_key = ?
		AND source_app_key = ?
		AND next_offset = ?`,
		n,
		ak,
		sk,
		c,
```

persistence/provider/sql/offsetstore.go
```diff
@@ -0,0 +1,100 @@
...
+		if ok, err := t.ds.driver.InsertOffset(
+			ctx,
+			t.actual,
+			t.ds.appKey,
+			ak,
+			n,
+		); ok || err != nil {
+			return err
+		}
+	}
```
jmalloc, line 80: It'd be easier to see at a glance what's going on here if the update/insert was pulled out into an `upsertOffset()` method and if `ok/err` check performed just once.

danilvpetrov: Should be good now. Спасибо за твое терпение, братан!

jmalloc: хаха, ты один из них, которых является терпеливым!

I know there's got to be a *way* more idiomatic way to say that, but I don't know what it is!

danilvpetrov:
you guys are awesome thanks very much for all your help!!!

PR #149, opened 2020-04-13
Improve naming of various "envelope" and "message" types.

### LINKED ISSUE #131: Log produced messages.

jmalloc:
Fixes #131

This PR changes the names of various "message container" types to give them distinct names.

The result is a structure as follows:

- The only type named `Message` is `dogma.Message`.
- `envelope.Envelope` is still the in-memory representation envelope, it contains a `Message`
- `envelopespec.Envelope` is still the on-disk/network representation of an envelope
- `Parcel` is used in various packages to refer to an envelope (of either kind) with additional meta-data that is specific to that package
- `Pair` is used in various packagers as a container for a `Parcel` and an `envelope.Envelope` - keeping the marshaled/unmarshaled forms together to avoid excess [un]marshaling.

In a separate branch/PR I'm going to experiment with removing the `envelope.Envelope` type altogether.


jmalloc:
See #152 instead.

PR #187, opened 2020-04-20
Serialize message handling for the same aggregate instance.

jmalloc:
This PR adds some coordination around the handling of messages that are routed to the same aggregate instance.

This is intended not only to reduce the likelihood of OCC conflicts and stalls when an aggregate instance receives a bursts of messages, but also to provide the necessary locking around an in-memory snapshot cache to be added in #185.

### DIFF
--- a/internal/x/syncx/namedmutex.go
+++ b/internal/x/syncx/namedmutex.go
@@ -1,0 +1,92 @@
+package syncx
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+)
+
+// UnlockFunc is a function used to unlock a previously locked mutex.
+type UnlockFunc func()
+
+// MutexNamespace is a "namespace" of named, context-aware mutexes.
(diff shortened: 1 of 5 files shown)


jmalloc:
This one's a bit like that `RWMutex`, the more eyes the better.

internal/x/syncx/namedmutex.go
```diff
@@ -0,0 +1,92 @@
...
+
+	ns.mutexes[n] = m
+
+	return m
+}
+
+// release decrements the locker count on m and removes it from ns.mutexes if
+// there are no other lockers.
+func (ns *MutexNamespace) release(n string, m *nmutex) {
+	// Remove outselves from the locker count. If the result is non-zero then
```
jmalloc, line 74:
```suggestion
	// Remove ourselves from the locker count. If the result is non-zero then
```

PR #192, opened 2020-04-21
Implement an in-memory cache of aggregate instances.

### LINKED ISSUE #185: Implement in-memory aggregate snapshot caching.

jmalloc:
Fixes #185

This PR introduces a highly-use-case-specific cache to the `aggregate.Sink`. Each handler has its own cache instance, and each record in the cache represents an aggregate instance.

Records are "acquired" from the cache, locking them so that no other goroutine can acquire them at the same time. This ensures that only one goroutine handles a message for a given instance at a time, reducing the likelihood of OCC failures without any IO. (This approach made #187 redundant, which is why I closed it).

When a cache record has been acquired, the acquirer must explicitly call a `KeepAlive()` method in order to retain that record in the cache when it's released. This approach feels a little unusual to me, but seems to work well in practice because it automatically invalidates the cache record whenever there is a failure condition, which for aggregates includes several very intentional `panic()`.  

Some other notes:

- I used `sync.Map`. I will need to benchmark to know that this is the right choice, but the intent is definitely that the cache should remain quite stable which suits `sync.Map` performance profile. This could definitely change to `map` + `mutex` with little issue.

- I tried to keep manipulations of the underlying map to a minimum. This is another request I chose the acquire/release approach, rather than get/set like you might normally have in a cache. When a record is manipulated and marked with `KeepAlive()` the same record simply stays in the map, there is no update of the map at all in the happy-path.

### DIFF
--- a/handler/aggregate/pipeline_test.go
+++ b/handler/aggregate/pipeline_test.go
@@ -74,6 +74,16 @@ var _ = Describe("type Sink", func() {
 
 		logger = &logging.BufferedLogger{}
 
+		handler.NewFunc = func() dogma.AggregateRoot {
+			return &AggregateRoot{
+				Value: &[]dogma.Message{},
+				ApplyEventFunc: func(m dogma.Message, v interface{}) {
+					p := v.(*[]dogma.Message)
+					*p = append(*p, m)
+				},
+			}
+		}
+
 		sink = &Sink{
 			Identity: &envelopespec.Identity{
 				Name: "<aggregate-name>",
@@ -192,8 +202,6 @@ var _ = Describe("type Sink", func() {
 		})
 
 		When("the instance exists", func() {
-			var root *AggregateRoot
-
 			BeforeEach(func() {
 				createReq, _ := NewPipelineRequestStub(
 					NewParcel("<created>", MessageC1),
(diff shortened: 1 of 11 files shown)


jmalloc:
As best I can tell the code changes are complete. The existing test suite is passing, after some modifications that make it care less about HOW the aggregate state is loaded, and rather just ensure that it has the right state.

The `aggregate.Sink` type needs explicit tests to ensure that the cache is invalidated after a failure, and the `cache` package itself has no tests yet.

handler/aggregate/pipeline_test.go
```diff
@@ -282,9 +283,24 @@ var _ = Describe("type Sink", func() {
...
+
+						// The root itself should also have been reset to as-new.
+						r := s.Root().(*AggregateRoot)
+						Expect(r.Value).To(Equal(
+							&[]dogma.Message{},
+						))
+
+						// We need to record an event whenever we create to
+						// prevent a panic, we do this after our assertion that
+						// the root is empty otherwise we this event applied.
```
koden-km, line 302: I'm not sure what you were aiming for here:
> otherwise we this event

The sentence could probably do with a full stop above also? "prevent a panic, we do this"

handler/cache/record.go
```diff
@@ -0,0 +1,92 @@
...
+		// tick we'll remove it.
+		r.state = idle
+	case idle:
+		// It's still idle, meaning it hasn't been acquired since
+		// the last tick.
+		r.remove()
+	}
+}
+
+// state is an enumeration that describes the records state in the cache.
```
danilvpetrov, line 85:
```suggestion
// state is an enumeration that describes the record state in the cache.
```

Issue #209, opened 2020-04-24
Decouple the pipeline package from the datastore.


jmalloc:
The`pipeline.Response` type currently handles persisting to the queue and event store directly. `queue.Queue` is notified about new messages via an observer. This logic could be much more direct, since it's not optional.

The `queue` package is already the one place that reads from the queue store, and it's already handling the writes for `CommandExecutor`, it makes sense to localise all of this related behavior.

The same thing will need to be done for the `eventbuffer`, or whatever the thing that #74 introduces ends up being called.

jmalloc:
The pipeline package no longer exists as of #251.

PR #210, opened 2020-04-24
Decouple `queue` package from `pipeline`.

jmalloc:
This PR decouples the `queue` package from `pipeline` by moving `queue.PipelineSource` to `pipeline.QueueStore`.

This is part of a broader effort to provide a clearer delineation between the "in memory" application coordination code and the persisted layers (see also #209).


jmalloc:
graph

You can see this further separation emerging in the modgraph.  It renders `handler/cache` in a weird spot, but just ignore that :P

Issue #74, opened 2020-03-28
Add a mechanism to wake `eventstream.PersistedStream` and the stream server.


jmalloc:
The idea is that when the engine saves new events to the store, it can wake any streams that are blocked waiting for a new event.

jmalloc:
Much more important as of #75.

PR #213, opened 2020-04-24
Implement `eventbuffer.Buffer`.

jmalloc:
Partially addresses #74 

This PR implements the "eventbuffer.Buffer" type, which is an in-memory buffer of the most recent events from a (local) application's event store that serves as a coordination point for multiple consumers waiting for new events to occur. 

It's analogous to `queue.Queue` which services a similar role for the message queue. However, whereas the `queue.Queue` type will eventually end up with the entire queue in memory as it shrinks, the `eventbuffer.Buffer` will only keep up to a fixed number of events as the store grows. I am open to better names than buffer!

Events are added to the buffer in contiguous batches, which is how they "emerge" from the pipeline observer after being committed in a transaction. However, the batches themselves can be out of order as each transaction is handled in a separate goroutine. To solve this the implementation uses a priority queue onto which it places batches of events that arrive out of order. In practice I would hope this "reordering" queue would remain very small, but I would ideally like to refactor to exclude this queue altogether.

The events from "recent history", once they have been reordered are stored in a fixed-size ring buffer implemented atop `[]*parcel.Parcel`.

Finally there is the `future` type which represents the "yet-to-arrive-ever-elusive-very-next-event".  Whenever a consumer requests an event at the next offset that is expected to occur, it waits upon this `future` which is fulfilled when that next event is added to the buffer.  The name `future` is taken by analogy to similar generic types in C# and C++, from memory, but it's not a concept we tend to see in Go.

### DIFF
--- a/eventstream/eventbuffer/buffer.go
+++ b/eventstream/eventbuffer/buffer.go
@@ -1,0 +1,258 @@
+package eventbuffer
+
+import (
+	"context"
+	"errors"
+	"sync"
+
+	"github.com/dogmatiq/dodeca/logging"
+	"github.com/dogmatiq/infix/internal/x/containerx/pqueue"
+	"github.com/dogmatiq/infix/parcel"
+	"github.com/dogmatiq/infix/persistence/subsystem/eventstore"
+)
+
+// DefaultSize is the default number of recent events to buffer in memory.
+const DefaultSize = 100
+
+// Buffer is an in-memory buffer of recently recorded events.
+type Buffer struct {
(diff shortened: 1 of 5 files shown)


jmalloc:
Test failures are unrelated, caused by #214

jmalloc:
FYI guys, I am making another attempt at this, taking a different approach, so please don't bother  reviewing this unless you particularly want to.

koden-km:
Were those recent changes your completed new approach?

jmalloc:
They're in #217

jmalloc:
This PR has been  superseded by #217.

PR #217, opened 2020-04-25
Reimplement a more serious memory stream.

jmalloc:
This is an alternative to #213, partially addresses #74. 

The approach taken here is to reimplement the existing `memorystream` to serve as the "recent history" buffer for the event store. The existing `memorystream` was a bit of a toy implementation suitable only for testing, and its utility in that role was somewhat diminished since the introduction of the `memory` persistence provider. It is still used for some tests after the changes in this PR.

This new implementation uses a singly-linked list of nodes containing the `[]*parcel.Parcel` slices that come directly from the `pipeline.Response`. Once the stream's internal buffer is initialised, historical reads are lock-free because the list nodes are immutable and the head/tail of the linked list can be obtained via an atomic operation, rather than a mutex lock.

I've retained much the same reordering approach as #213, using a priority queue.

I have some doubts about using a node-per-slice given that for most applications it's highly likely that each slice will usually contain a single parcel. Since the intent is that consumers will first look directly in the event store and then only hit this implementation once they are at or near the end they should obtain a reference to a relevant node quickly and not have to scan through a large number of nodes to find their starting offset.

This PR also introduces the `eventstream.ErrTruncated` error used to indicate that the stream can not provide access to old events.


jmalloc:
I've gone a bit silly in this PR; doing things for the sake of "performance" I wouldn't normally do without benchmarking. It's been a bit of a learning experience, but I could well be doing things that are slower than more "naive" approaches so that benchmarking is still going to have to happen.

eventstream/memorystream/node.go
```diff
@@ -0,0 +1,114 @@
...
+	// If we can't do this it means that a call to acquire() occured after we
+	// decremented the reference count but before we attempted this CAS.
+	if atomic.CompareAndSwapInt32(&n.refs, 0, math.MinInt32) {
+		pooling.ParcelSlices.Put(n.parcels)
+	}
+}
+
+// advance returns the next node in the list.
+//
+// It blocks until the next node is linked or or ctx is canceled.
```
koden-km, line 56:
```suggestion
// It blocks until the next node is linked or ctx is canceled.
```

eventstream/memorystream/stream.go
```diff
@@ -48,48 +64,199 @@ func (s *Stream) Open(
...
+		// to load the head and step through the list until we find the node
+		// that contains the offset we want.
+
+		n.release() // release the tail node
+
+		n = s.loadHead()
+		for !n.acquire() {
+			// CODE COVERAGE: The head node we *just* loaded has been
+			// invalidated and replaced with a new head node. This hard to time
+			// correctly in a test, but likely to occur in production.
```
koden-km, line 108:
```suggestion
			// invalidated and replaced with a new head node. This is hard to
			// time correctly in a test, but likely to occur in production.
```

eventstream/memorystream/stream.go
```diff
@@ -48,48 +64,199 @@ func (s *Stream) Open(
...
-	for _, p := range parcels {
-		t := message.TypeOf(p.Message)
-		if !s.Types.Has(t) {
-			panic("unsupported message type: " + t.String())
+	// Regardless of whether these events need to be reordered we include them
+	// in the size for the purpose of enforcing the size limit.
+	s.size += len(n.parcels)
+
+	if n.begin > tail.end {
+		// These events has arrived out of order, keep the node in the reorder
```
koden-km, line 164:
```suggestion
		// These events have arrived out of order, keep the node in the reorder
```

eventstream/memorystream/stream.go
```diff
@@ -3,24 +3,40 @@ package memorystream
...
 
-	m      sync.Mutex
-	ready  chan struct{}
-	events []*eventstream.Event
+	// FirstOffset is the first offset that will be kept in this stream.
+	// This is used to determine whether Open() should return ErrTruncated.
+	FirstOffset eventstream.Offset
+
+	// BufferSize is the maximum number of messages to buffer in memory.
+	// If it is non-positive, DefaultBufferSize is used.
```
danilvpetrov, line 32: What if it is zero?

jmalloc:
I've backed out some of the more egregious premature optimisations - starting with removing the pooling of slices, this has a knock-on effect that allowed removal of some manual reference counting which was only used to determine when a slice could be returned to the pool.

eventstream/memorystream/node.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// acquire "locks" the node for reading.
+func (n *node) acquire() bool {
+	// Increment the reference count. It should swing positive. If it doesn't it
+	// means that this node has already been invalidated.
+	return atomic.AddInt32(&n.refs, 1) > 0
+}
+
+// release "unlocks" a previously call to acquire().
```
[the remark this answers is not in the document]
jmalloc: This method is gone now.

eventstream/memorystream/node.go
jmalloc: This method is gone now.

eventstream/memorystream/stream.go
jmalloc: Fixed.

eventstream/memorystream/stream.go
jmalloc: Zero is non-positive :D

eventstream/memorystream/stream.go
jmalloc: This logic is gone now.

eventstream/memorystream/stream.go
```diff
@@ -48,48 +95,202 @@ func (s *Stream) Open(
...
+	s.size += len(n.parcels)
+
+	if n.begin > tail.end {
+		// These events have arrived out of order, keep the node in the reorder
+		// queue until it can be linked to its preceeding node.
+		s.reorder.Push(n)
+		return
+	}
+
+	// These events are the ones we were expecting next. First, we we make the
```
koden-km, line 203:
```suggestion
	// These events are the ones we were expecting next. First, we make the
```

eventstream/memorystream/stream.go
```diff
@@ -48,48 +95,202 @@ func (s *Stream) Open(
...
 
-	if s.ready != nil {
-		close(s.ready)
-		s.ready = nil
+	// Load the current head.
+	head := s.loadHead()
+
+	// Continue to replace head with head.next until we're under the limit or we
+	// reach the tail. We can read head.next here without using node.next()
+	// because we still hold the lock on the mutex any node.link() calls have
```
koden-km, line 248: I think this sentence needs a comma or full stop and start a new sentence?

jmalloc:
Superseded by #233

PR #219, opened 2020-04-25
Refactor `pipeline` package to only allow single observers.

jmalloc:
These observers are not optional, one is always the `queue.Queue` and the other will be the `memorystream` once #74 is complete.

This PR reworks the `pipeline` package to only allow one of each of these observers and to make it clear that the observers take ownership of the slices used to pass messages around.

To some small extent this also partially addresses #210.


jmalloc:
Man, I have no idea wth Codecov is complaining about at this point, these tests are driving me nuts.

PR #230, opened 2020-04-27
Use memory stream as a cache of recent events.

### LINKED ISSUE #74: Add a mechanism to wake `eventstream.PersistedStream` and the stream server.

jmalloc:
Blocked by #217, fixes #74

This PR changes `persistedstream.Stream` to use a `memorystream.Stream` as a cache of recent events.

`persistedstream.cursor` now hits this cache first, falling back to query an `eventstore.Repository` if the required offset is too old to be in the cache.  

Once the cursor "runs out" of events in the store, in other words, it has caught up to the tail of the stream, it tries to "lock onto" the memory stream again, using it exclusively from then on.

### DIFF
--- a/app.go
+++ b/app.go
@@ -10,6 +10,7 @@ import (
 	"github.com/dogmatiq/dogma"
 	"github.com/dogmatiq/infix/draftspecs/envelopespec"
 	"github.com/dogmatiq/infix/eventstream"
+	"github.com/dogmatiq/infix/eventstream/memorystream"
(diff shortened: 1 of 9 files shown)


jmalloc:
I've added you guys as reviewers since this is the logical conclusion to #233 that you guys reviewed for me and Kev was asking questions about aspects that are covered by this PR, but if you don't have time to look it over, I'm far less worried about these changes :)

Issue #220, opened 2020-04-26
Improve relationship between aggregate instances and their events.


jmalloc:
~Blocked by #227~

### Summary

This change is about inverting the way aggregate instances are related to the events they produce in the persistence layer.

**In short, rather than storing event ranges in aggregate instance meta-data we will store aggregate instance revisions in the event store.**

### Rationale

The `SaveOffset()` operation is currently the only transactional operation that needs to return information to the caller (the event offset). Furthermore, the only reason it needs to do this is so that the offsets in aggregate meta-data can be populated to handle the notion of "instance destruction".

There are a few reasons why returning data from a transaction operation  is undesirable:

- The SQL provider implementations have to lock the `event_offset` table when the first event is recorded until transaction is committed.
- It makes it difficult to batch recording of multiple events in a single database operation.
- Not all persistence providers would be able to implement this easily or at all. DynamoDB (#175) for instance, supports transactions but they are submitted as a singular operation via a HTTP API.

### Proposal

As it stands, `aggregatestore.MetaData` contains an event offset range that defines the interval that includes the instance's historical events. When an instance is destroyed the lower-bound of that range is updated to the upper-bound, forming a 0-length range after all existing events.

~What I'm proposing is that we instead store the instance's revision against each event in the event store, and query based on the revision that the instance was last destroyed. The event store already persists the handler key and instance ID, so this is not some new coupling of systems that did not exist before.~

This obviates the need for the offset return value, and potentially opens us up to submitting events to the database in a single round-trip.

**Edit:** After talking with Danil, we've decided to use the "message ID" as a sentinel to work out which events occurred after the last aggregate deletion. This avoids any coupling between the "eventstore" and the "queuestore" altogether.
### Checklist

These are items that I think we'll need to address directly as part of this issue. 

- [ ] ~Work out how to pass the revision to `SaveEvents()`, some ideas:~
    - ~**Add a separate `SaveEventForAggregate(rev, envelope)` method**~
    - ~Use functional options `(SaveEvent(env, WithAggregateRevision(rev))`~
    - ~Pass an options struct~
- [x] Allow querying of events by aggregate revision, options
    - Add `MinAggregateRevision` to query
    - **Add separate `LoadEventsBySource()` method** 
- [x] Replace offset range in `aggregatestore.MetaData` with a ~revision range~ `DeletedByMessageID`.
- [x] Change aggregate loader to use `LoadEventsBySource()` & drop aggregate related fields from `Query`

[...]

Issue #227, opened 2020-04-27
Add a `TransactionResult` type.


jmalloc:
The transaction result would be returned from `Transaction.Commit()` on success, and include the offset of the first event saved within this transaction, if any.

jmalloc:
Rather than putting the offset in this new result type, we should place the same "item" slices that appear in `pipeline.Response`, that will help with #210 and eventually they can be removed from pipeline.Response.

danilvpetrov:
So generally the Transaction interface inside `persistence` pkg would look like the following (including the new `TransactionResult` type):

```go
// Transaction exposes persistence operations that can be performed atomically.
// Transactions are not safe for concurrent use.
type Transaction interface {
	aggregatestore.Transaction
	eventstore.Transaction
	queuestore.Transaction
	offsetstore.Transaction

	// Commit applies the changes from the transaction.
	Commit(ctx context.Context) (TransactionResult, error)

	// Rollback aborts the transaction.
	Rollback() error
}

// TransactionResult is the result returned by Transaction.Commit() method.
type TransactionResult struct {
	EventItems []*eventstore.Item
	QueueItems []*queuestore.Item
}
```
Is this what you're thinking?

jmalloc:
Yep, exactly!

danilvpetrov:
Fixed by  #235.

PR #231, opened 2020-04-27
Remove custom integer types for offsets and revisions.

### LINKED ISSUE #172: Simplify the various different "offset" types.
As the persistence system grows more complex and more subsystems make references to events within the store it becomes increasingly nonsensical to have separate `Offset` types in each package.

This idea was originally intended to distinguish between events that were definitely persisted in the store (`eventstore.Offset`) and events that may belong to some other application, having been consumed via a stream (`eventstream.Offset`).

I now have a requirement to reference offsets of persisted events in the `aggregatestore` package. I don't particularly like the idea of referring to `eventstore.Offset` from `aggregatestore`, nor do I like the idea of defining yet another `aggregatestore.Offset` type. The conversions between all the types defeat the safety they were originally intended to provide.

I think we need to return to only having two types. For persisted offsets, I think it makes sense to move `eventstore.Offset` to `persistence.EventOffset`, and share it across all persistence subsystems.

I'm not so sure what to do with `eventstream.Offset`, however. This is the type that should be used from the `offsetstore` ideally. However, my gut feeling is that there's something incorrect if we're making the persistence layer depend on this `eventstream`. At time of writing in `master` no persistence subsystems depend on anything outside the persistence layer except the `envelopespec` package.

Also, because `eventstream` has a stream implementation that's based on the `eventstore`, it would create a package dependency chain that looks like `persistence/offsetstore` -> `eventstream` -> `persistence/eventstore`, sort of "leaving" the persistence layer and "returning" to it, which also makes me uncomfortable. 

/cc @danilvpetrov this is relevant to your `offsetset` store PR #148.

jmalloc:
Fixes #172 

Ok, I bit the bullet and just did this. Like I was saying, all the casting between the offset types (and `uint64` from protobuf) really diminished the utility of these types and they just contributed to tighter coupling between packages than we really necessary.

I've also removed the `Revision` types while I was at it, because as on #220 they would have contributed the same kind of problem.

These graphs can be a bit misleading, but just to illustrate the coupling between packages, here is the graph from `master`:

graph-master

And the graph as of this PR:

graph

### DIFF
--- a/eventstream/consumer.go
+++ b/eventstream/consumer.go
@@ -17,13 +17,13 @@ type Handler interface {
 	// specific application's event stream.
 	//
 	// id is the identity of the source application.
-	NextOffset(ctx context.Context, id configkit.Identity) (Offset, error)
+	NextOffset(ctx context.Context, id configkit.Identity) (uint64, error)
 
 	// HandleEvent handles an event obtained from the event stream.
 	//
 	// o must be the offset that would be returned by NextOffset(). On success,
 	// the next call to NextOffset() will return ev.Offset + 1.
-	HandleEvent(ctx context.Context, o Offset, ev *Event) error
+	HandleEvent(ctx context.Context, o uint64, ev *Event) error
 }
 
 // Consumer reads events from a stream in order to handle them.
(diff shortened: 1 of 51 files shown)


jmalloc:
@danilvpetrov I thought I'd try to get this in before your started on #227

PR #232, opened 2020-04-27
Fix violations reported by golangci-lint.

jmalloc:
This PR is the result of fixing the violations reported by [golangci-lint](https://github.com/golangci/golangci-lint). I'm considering adding this to the Dogmatiq CI configuration as I've noticed a few things cropping up here and there that would be good to catch earlier.

For this PR, I ran the linter as follows:

```
golangci-lint run --exclude SA1019   
```

`SA1019` is just a warning from the `staticcheck` linter about the deprecated protobuf package.

There's a tonne of annoying "false positives" about unchecked errors that I've had to silence with a special `// nolint` comment. In general I've tried to actually check or make use of the error value if possible, but where it's not I've adopted some conventions about how to do the silencing. 

One thing that's annoying is that these errors can come from one of two different linters:`errcheck` and `staticcheck`.

Anyway, these are the conventions I adopted:

- `go` and `defer` statements that don't check error values can be silenced with `// nolint`. No further explanation is necessary, this is common and unfortunate.
- Other errors that don't need to be checked are silenced with `// nolint:errcheck` specifically. If they're in the tests that's enough, but for any "real" code I've included an additional explanation like so: `// nolint:errcheck // (explanation commit)`

### DIFF
--- a/app.go
+++ b/app.go
@@ -160,7 +160,7 @@ func (e *Engine) newPipeline(
 		},
 	}
 
-	if err := cfg.AcceptRichVisitor(nil, rf); err != nil {
+	if err := cfg.AcceptRichVisitor(context.Background(), rf); err != nil {
 		panic(err)
 	}
 
--- a/app.go
+++ b/app.go
@@ -160,7 +160,9 @@ func (e *Engine) newPipeline(
 		},
 	}
 
-	cfg.AcceptRichVisitor(nil, rf)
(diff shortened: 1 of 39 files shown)


jmalloc:
As per Slack discussion with @danilvpetrov, we don't really want to take this on atm. Having to suppress so many warnings makes the tool a bit redundant. I've removed the `// nolint` comments, but will retain the fixes and just run this manually now and then.

PR #233, opened 2020-04-28
memorystream with single-event linked list nodes

jmalloc:
This PR is yet another alternative to #217 that uses `eventstore.Event` to represent the events and only stores a single event per linked-list node.

I just wanted to try this approach given that the vast majority of "batches" would have had a single event in them (one command in, one event out, typically). It simplifies the implementation and removes a few edge cases that were present before, such as 0-length batches used as sentinel nodes.


eventstream/memorystream/stream.go
```diff
@@ -48,48 +89,200 @@ func (s *Stream) Open(
...
 		offset: o,
-		closed: make(chan struct{}),
 		filter: f,
+		node:   n,
+		closed: make(chan struct{}),
 	}, nil
 }
 
-// Append appends events to the stream.
+// Append adds events to the the tail of the stream.
```
koden-km, line 128:
```suggestion
// Append adds events to the tail of the stream.
```

eventstream/memorystream/stream.go
```diff
@@ -48,48 +89,200 @@ func (s *Stream) Open(
...
+
+	if ev.Offset < tail.offset {
+		// This event is earlier than what we're expecting next so we just
+		// discard it. This "should never happen" provided that s.FirstOffset is
+		// populated before new events are allowed to be produced when the
+		// engine first starts.
+		return
+	}
+
+	// Regardless of whether this event need to be reordered or not we include
```
koden-km, line 184:
```suggestion
	// Regardless of whether this event needs to be reordered or not we include
```

eventstream/memorystream/stream_test.go
```diff
@@ -35,20 +38,168 @@ var _ = Describe("type Stream", func() {
...
+
+		When("the maximum buffer size is exceeded", func() {
+			It("drops old events when the max buffer size is reached", func() {
+				stream.BufferSize = 3
+
+				addEvents(stream, 100, 101)
+				addEvents(stream, 102, 103)
+
+				expectEventsToBeTruncated(ctx, stream, 100, 100)
+				expectEventsToBeAvailable(ctx, stream, 102, 103)
```
koden-km, line 122: This isn't wrong, but should this also be expecting `101` to be available? A left over from your old batch version of this PR?

eventstream/memorystream/stream_test.go
```diff
@@ -35,20 +38,168 @@ var _ = Describe("type Stream", func() {
...
+			})
+
+			It("drops the oldest events if a single call exceeds the buffer size", func() {
+				stream.BufferSize = 3
+
+				addEvents(stream, 100, 103)
+				expectEventsToBeTruncated(ctx, stream, 100, 100)
+				expectEventsToBeAvailable(ctx, stream, 102, 103)
```
koden-km, line 130: Same comment as above.

eventstream/memorystream/node.go
```diff
@@ -0,0 +1,71 @@
...
+	if ptr == nil {
+		ch := make(chan struct{})
+		ptr = unsafe.Pointer(&ch)
+
+		if !atomic.CompareAndSwapPointer(&n.resolved, nil, ptr) {
+			// CODE COVERAGE: Another goroutine managed to initialize the
+			// channel since we called LoadPointer(), which is hard to time
+			// correctly in tests.
+			//
+			// Note, we have to use the the existing channel otherwise at least
```
danilvpetrov, line 36:
```suggestion
			// Note, we have to use the existing channel otherwise at least
```

eventstream/memorystream/stream.go
```diff
@@ -48,48 +89,200 @@ func (s *Stream) Open(
...
+	if n == nil {
+		// If the tail is nil then the list is uninitialized. We initialize it
+		// with a node that will contain the as-yet-unknown event at
+		// c.FirstOffset.
+		s.m.Lock()
+		n = s.init()
+		s.m.Unlock()
+	} else if o < n.offset {
+		// We already have an initialized list, and the requested offset is for
+		// an historical event. Starting at the head, the cursor will scan
```
danilvpetrov, line 115:
```suggestion
		// a historical event. Starting at the head, the cursor will scan
```

jmalloc: Well, look at you my "I talk properly and actually pronounce the H"! :D

Issue #168, opened 2020-04-17
Remove offset return value from `SaveEvent()`.


jmalloc:
Blocked by #220 

This was originally added so that the aggregate system could use the offset to enforce OCC for aggregate instance, but since each aggregate instance now has its own revision this is no longer necessary.

I create this issue in the interest of keeping the persistence layer as slim as possible - limited to the exact functionality we need - so that it's easier to add new providers.

jmalloc:
Blocked by #171

jmalloc:
This is definitely still needed.

jmalloc:
Reopening to attempt to remove this once again.

PR #235, opened 2020-04-28
Add `TransactionResult` type.

danilvpetrov:
This PR adds `persistence.TransactionResult` type returned by `persistence.Transaction.Commit()` method. This type is the container for all committed event for the committed event and queue items that looks like the following:
```go
// TransactionResult is the result returned by Transaction.Commit() method.
type TransactionResult struct {
	EventItems []*eventstore.Item
	QueueItems []*queuestore.Item
}
```

### DIFF
--- a/persistence/provider/boltdb/transaction.go
+++ b/persistence/provider/boltdb/transaction.go
@@ -34,7 +34,7 @@ func (t *transaction) Commit(
 		return &t.result, t.actual.Commit()
 	}
 
-	return t.result, nil
+	return &t.result, nil
 }
 
 // Rollback aborts the transaction.
--- a/persistence/provider/boltdb/transaction.go
+++ b/persistence/provider/boltdb/transaction.go
@@ -11,7 +11,7 @@ import (
 // data stores.
 type transaction struct {
 	ds     *dataStore
-	result *persistence.TransactionResult
+	result persistence.TransactionResult
 	appKey []byte
 	actual *bbolt.Tx
 }
@@ -31,10 +31,7 @@ func (t *transaction) Commit(
 	}
 
 	if t.actual != nil {
-		if err := t.actual.Commit(); err != nil {
-			return nil, err
-		}
-
+		return &t.result, t.actual.Commit()
 	}
 
 	return t.result, nil
@@ -74,8 +71,6 @@ func (t *transaction) begin(ctx context.Context) error {
 		t.actual = t.ds.db.Begin(ctx)
 	}
 
-	t.result = &persistence.TransactionResult{}
-
 	return nil
 }
 
@@ -89,5 +84,4 @@ func (t *transaction) end() {
 	}
 
 	t.ds = nil
-	t.result = nil
 }
--- a/persistence/provider/boltdb/transaction.go
+++ b/persistence/provider/boltdb/transaction.go
@@ -11,27 +11,33 @@ import (
 // data stores.
 type transaction struct {
 	ds     *dataStore
+	result *persistence.TransactionResult
 	appKey []byte
 	actual *bbolt.Tx
 }
 
 // Commit applies the changes from the transaction.
-func (t *transaction) Commit(ctx context.Context) error {
+func (t *transaction) Commit(
+	ctx context.Context,
+) (*persistence.TransactionResult, error) {
 	defer t.end()
 
 	if t.ds == nil {
-		return persistence.ErrTransactionClosed
+		return nil, persistence.ErrTransactionClosed
 	}
 
 	if err := t.ds.checkOpen(); err != nil {
-		return err
+		return nil, err
 	}
 
 	if t.actual != nil {
-		return t.actual.Commit()
+		if err := t.actual.Commit(); err != nil {
+			return nil, err
+		}
+
 	}
 
-	return nil
+	return t.result, nil
 }
 
 // Rollback aborts the transaction.
@@ -68,6 +74,8 @@ func (t *transaction) begin(ctx context.Context) error {
 		t.actual = t.ds.db.Begin(ctx)
 	}
 
+	t.result = &persistence.TransactionResult{}
+
 	return nil
 }
 
@@ -81,4 +89,5 @@ func (t *transaction) end() {
 	}
 
 	t.ds = nil
+	t.result = nil
 }
--- a/persistence/provider/boltdb/transaction.go
+++ b/persistence/provider/boltdb/transaction.go
@@ -19,22 +19,23 @@ type transaction struct {
 // Commit applies the changes from the transaction.
 func (t *transaction) Commit(
 	ctx context.Context,
-) (*persistence.TransactionResult, error) {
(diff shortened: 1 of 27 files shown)


persistence/transaction_test.go
```diff
@@ -63,10 +65,61 @@ var _ = Describe("func WithTransaction", func() {
 		Expect(items).NotTo(BeEmpty())
 	})
 
+	It("returns the transaction result", func() {
```
jmalloc, line 68: IMO this test should be moved into the `providertest/eventstore` package.

```suggestion
	It("includes event store items in the transaction result", func() {
```

persistence/transaction.go
```diff
@@ -23,12 +23,19 @@ type Transaction interface {
...
 
 	// Commit applies the changes from the transaction.
-	Commit(ctx context.Context) error
+	Commit(ctx context.Context) (*TransactionResult, error)
 
 	// Rollback aborts the transaction.
 	Rollback() error
 }
 
+// TransactionResult is the result returned by Transaction.Commit() method.
```
jmalloc, line 32:
```suggestion
// TransactionResult contains information about a successfully committed transaction.
```

persistence/transaction.go
```diff
@@ -23,12 +23,19 @@ type Transaction interface {
...
+type TransactionResult struct {
+	// EventItems is the slice of eventstore.Item persisted within the boundary
+	// of the transaction.
```
jmalloc, line 35:
```suggestion
	// EventItems contains the events persisted within the transaction.
```

persistence/provider/boltdb/transaction.go
```diff
@@ -11,27 +11,33 @@ import (
 // data stores.
 type transaction struct {
 	ds     *dataStore
+	result *persistence.TransactionResult
```
jmalloc, line 14:
```suggestion
	result persistence.TransactionResult
```

Regardless of whether we choose to retain the pointer return value from `Commit()`, these `result` fields don't need to be a pointer - they're always needed in the happy path.

persistence/provider/boltdb/transaction.go
```diff
@@ -11,27 +11,33 @@ import (
...
-		return err
+		return nil, err
 	}
 
 	if t.actual != nil {
-		return t.actual.Commit()
+		if err := t.actual.Commit(); err != nil {
+			return nil, err
+		}
+
```
jmalloc, line 37: This would get rid of those those uncoverable lines. I wouldn't recommend that if it was something we could reproduce in any sane way or we were responsible for generating the error in our code.

```suggestion
		return &t.result, t.actual.Commit()
```

persistence/provider/memory/eventstore.go
```diff
@@ -19,7 +20,11 @@ func (t *transaction) SaveEvent(
 		return 0, err
 	}
 
-	return t.event.stageSave(&t.ds.db.event, env), nil
+	t.result.EventItems = append(
+		t.result.EventItems,
+	)
+
```
jmalloc, line 26: Probably a remnant of an earlier approach.

```suggestion
	t.result.EventItems = append(
		t.result.EventItems,
	)

```

persistence/provider/memory/eventstore.go
```diff
@@ -126,6 +132,8 @@ func (cs *eventStoreChangeSet) stageSave(
 
 	cs.items = append(cs.items, item)
 
+	result.EventItems = append(result.EventItems, item)
```
jmalloc, line 135: I think you can handle this entirely in `Commit()`. You have access to the `eventStoreChangeSet` there, so you can just use `cs.items` as `EventItems` in the result. There are a few benefits to doing so:

- this code wont need to change when `SaveEvent()` loses it's offset return value (#168 
- avoid append()'ing
- don't need `transaction.result` field anymore

persistence/provider/sql/transaction.go
```diff
@@ -12,25 +12,30 @@ import (
...
 	if err := t.ds.checkOpen(); err != nil {
-		return err
+		return nil, err
 	}
 
 	if t.actual != nil {
-		return t.actual.Commit()
+		if err := t.actual.Commit(); err != nil {
+			return nil, err
+		}
```
jmalloc, line 35: As per suggestion for BoltDB:

```suggestion
		return &t.result, t.actual.Commit()
```

persistence/provider/sql/transaction.go
```diff
@@ -12,25 +12,30 @@ import (
 type transaction struct {
 	ds     *dataStore
 	actual *sql.Tx
+	result *persistence.TransactionResult
```
jmalloc, line 15:
```suggestion
	result persistence.TransactionResult
```

persistence/provider/boltdb/transaction.go
```diff
@@ -68,6 +74,8 @@ func (t *transaction) begin(ctx context.Context) error {
 		t.actual = t.ds.db.Begin(ctx)
 	}
 
+	t.result = &persistence.TransactionResult{}
+
```
jmalloc, line 78:
```suggestion
```

No longer necessary when not a pointer.

persistence/provider/boltdb/transaction.go
```diff
@@ -81,4 +89,5 @@ func (t *transaction) end() {
 	}
 
 	t.ds = nil
+	t.result = nil
```
jmalloc, line 92:
```suggestion
```

No longer necessary when not a pointer.

persistence/provider/memory/eventstore.go
danilvpetrov: Cool, i will do this in a follow-up commit.

persistence/provider/memory/eventstore.go
danilvpetrov: Hmm, is it one of those suggestions that are the same as the original, GH would not probably like that.

jmalloc: Sorry man! Fixed it. I think there was an extra newline after or something too.

persistence/provider/memory/transaction.go
```diff
@@ -18,27 +18,31 @@ type transaction struct {
...
 	}
 
 	if err := t.ds.checkOpen(); err != nil {
-		return err
+		return nil, err
 	}
 
 	if !t.hasLock {
-		return nil
+		return nil, nil
```
jmalloc, line 35:
```suggestion
		return &persistence.TransactionResult{}, nil
```

This is a transaction in which nothing happened, so it still needs a valid result, it's just empty. This means we're missing a test-case that an empty `TransactionResult` is returned when no operations are done in the transaction. That should go in `providertest/transaction.go`.

danilvpetrov: I would assume that this change will be required not only for in-memory engine, right?

jmalloc: I think it's only this one, but just add the test and you'll know :D

PR #244, opened 2020-05-04
Add LoadEventsForAggregate() method.

danilvpetrov:
This PR partially addressed #220 and adds `LoadEventsForAggregate()` to the `evenstore.Repository` type.

### DIFF
--- a/fixtures/eventstore.go
+++ b/fixtures/eventstore.go
@@ -11,7 +11,26 @@ import (
 type EventStoreRepositoryStub struct {
 	eventstore.Repository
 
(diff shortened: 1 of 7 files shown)


danilvpetrov:
I am not sure if it's a good idea to rely on `eventstore.Query.AggregateHandlerKey` and `eventstore.Query.AggregateInstanceID` Fields in this PR. As these fields are supposed to disappear within the scope of #220, I was hoping for a bigger refactor PR that would create aggregate-only analogue of `eventstore.Query`. Pls LMK if we need this refactoring in this PR.

PR #253, opened 2020-05-06
Add LoadEventsBySource() method in eventstore.Repository

danilvpetrov:
This PR partially addresses #220 by adding `LoadEventsBySource()` method to `eventstore.Repository` and modifying the implementations of `eventstore.Transaction.SaveEvent()` method to persist message ID in a separate field (if not doing that already) so that `eventstore.Repository.LoadEventsForAggregate()` method can load source events from the given message ID.

### DIFF
--- a/persistence/internal/providertest/eventstore/repository.go
+++ b/persistence/internal/providertest/eventstore/repository.go
@@ -204,9 +204,9 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 			})
 		})
 
-		ginkgo.Describe("func LoadEventsForAggregate()", func() {
+		ginkgo.Describe("func LoadEventsBySource()", func() {
 			ginkgo.It("returns an empty result if the store is empty", func() {
-				items := loadEventsForAggregate(tc.Context, repository, "<aggregate>", "<instance-a>", "")
+				items := loadEventsBySource(tc.Context, repository, "<aggregate>", "<instance-a>", "")
 				gomega.Expect(items).To(gomega.BeEmpty())
 			})
 
@@ -225,7 +225,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 					item5.Envelope,
 				)
 
-				items := loadEventsForAggregate(
+				items := loadEventsBySource(
 					tc.Context,
 					repository,
 					"<aggregate>",
@@ -241,7 +241,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 					)
 				}
 
-				items = loadEventsForAggregate(
+				items = loadEventsBySource(
 					tc.Context,
 					repository,
 					"<aggregate>",
@@ -273,7 +273,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 					item5.Envelope,
 				)
 
-				items := loadEventsForAggregate(
+				items := loadEventsBySource(
 					tc.Context,
 					repository,
 					"<aggregate>",
@@ -289,7 +289,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 					)
 				}
 
-				items = loadEventsForAggregate(
+				items = loadEventsBySource(
 					tc.Context,
 					repository,
 					"<aggregate>",
@@ -307,7 +307,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 			})
 
 			ginkgo.It("does not return an error if events exist beyond the end offset", func() {
-				res, err := repository.LoadEventsForAggregate(tc.Context, "<aggregate>", "<instance-a>", "")
+				res, err := repository.LoadEventsBySource(tc.Context, "<aggregate>", "<instance-a>", "")
 				gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
 				defer res.Close()
 
@@ -326,7 +326,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 			})
 		})
 
-		ginkgo.Describe("func QueryEvents() and LoadEventsForAggregate()", func() {
+		ginkgo.Describe("func QueryEvents() and loadEventsBySource()", func() {
 			ginkgo.It("allows concurrent consumers for the same application", func() {
 				saveEvents(
 					tc.Context,
@@ -357,7 +357,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 				fn2 := func() {
 					defer g.Done()
 					defer ginkgo.GinkgoRecover()
-					items := loadEventsForAggregate(
+					items := loadEventsBySource(
 						tc.Context,
 						repository,
 						"<aggregate>",
--- a/persistence/internal/providertest/eventstore/repository.go
+++ b/persistence/internal/providertest/eventstore/repository.go
@@ -210,7 +210,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 				gomega.Expect(items).To(gomega.BeEmpty())
 			})
 
-			ginkgo.It("it returns a result containing the events that match the aggregate instance", func() {
+			ginkgo.It("returns a result containing the events that match the aggregate instance", func() {
 				expectedA := []*eventstore.Item{item0, item2}
 				expectedB := []*eventstore.Item{item1, item3}
 
@@ -230,7 +230,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 					repository,
 					"<aggregate>",
 					"<instance-a>",
-					item0.Envelope.MetaData.MessageId,
+					"",
 				)
 
 				for i, item := range items {
@@ -246,7 +246,55 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 					repository,
 					"<aggregate>",
 					"<instance-b>",
-					item1.Envelope.MetaData.MessageId,
+					"",
+				)
+
+				for i, item := range items {
+					expectItemToEqual(
+						item,
+						expectedB[i],
+						fmt.Sprintf("item at index #%d of slice", i),
+					)
+				}
+			})
+
+			ginkgo.It("matches the events for the aggregate only from the given message id, if specified ", func() {
+				expectedA := []*eventstore.Item{item2}
+				expectedB := []*eventstore.Item{item3}
+
+				saveEvents(
+					tc.Context,
+					dataStore,
+					item0.Envelope,
+					item1.Envelope,
+					item2.Envelope,
+					item3.Envelope,
+					item4.Envelope,
+					item5.Envelope,
+				)
+
+				items := loadEventsForAggregate(
+					tc.Context,
+					repository,
+					"<aggregate>",
+					"<instance-a>",
+					item2.Envelope.MetaData.MessageId,
+				)
+
+				for i, item := range items {
+					expectItemToEqual(
+						item,
+						expectedA[i],
+						fmt.Sprintf("item at index #%d of slice", i),
+					)
+				}
+
+				items = loadEventsForAggregate(
+					tc.Context,
+					repository,
+					"<aggregate>",
+					"<instance-b>",
+					item3.Envelope.MetaData.MessageId,
 				)
 
 				for i, item := range items {
--- a/persistence/internal/providertest/eventstore/repository.go
+++ b/persistence/internal/providertest/eventstore/repository.go
@@ -321,6 +321,13 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 					item2.Envelope.MetaData.MessageId,
 				)
 				gomega.Expect(err).Should(gomega.HaveOccurred())
+				gomega.Expect(err).To(
+					gomega.MatchError(
+						"message with ID '" +
+							item2.Envelope.MetaData.MessageId +
+							"' cannot be found",
+					),
+				)
 			})
 
 			ginkgo.It("does not return an error if events exist beyond the end offset", func() {
--- a/persistence/internal/providertest/eventstore/repository.go
+++ b/persistence/internal/providertest/eventstore/repository.go
@@ -258,7 +258,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 				}
 			})
 
-			ginkgo.It("matches the events from the source only after the given message id", func() {
+			ginkgo.It("only matches the events after the barrier message", func() {
 				expectedA := []*eventstore.Item{item2}
 				expectedB := []*eventstore.Item{item3}
 
@@ -306,7 +306,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 				}
 			})
 
-			ginkgo.It("returns an error if the message with the given id is not found", func() {
+			ginkgo.It("returns an error if the barrier message is not found", func() {
 				saveEvents(
 					tc.Context,
 					dataStore,
--- a/persistence/internal/providertest/eventstore/repository.go
+++ b/persistence/internal/providertest/eventstore/repository.go
@@ -210,7 +210,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 				gomega.Expect(items).To(gomega.BeEmpty())
 			})
 
-			ginkgo.It("returns a result containing the events that match the aggregate instance", func() {
+			ginkgo.It("returns a result containing the events that match the source instance", func() {
 				expectedA := []*eventstore.Item{item0, item2}
 				expectedB := []*eventstore.Item{item1, item3}
 
@@ -258,7 +258,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 				}
 			})
 
-			ginkgo.It("matches the events for the aggregate only from the given message id, if specified ", func() {
+			ginkgo.It("matches the events from the source only after the given message id", func() {
 				expectedA := []*eventstore.Item{item2}
 				expectedB := []*eventstore.Item{item3}
 
@@ -278,7 +278,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 					repository,
 					"<aggregate>",
 					"<instance-a>",
-					item2.Envelope.MetaData.MessageId,
+					item0.Envelope.MetaData.MessageId,
 				)
 
 				for i, item := range items {
@@ -294,7 +294,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 					repository,
 					"<aggregate>",
 					"<instance-b>",
-					item3.Envelope.MetaData.MessageId,
+					item1.Envelope.MetaData.MessageId,
 				)
 
 				for i, item := range items {
@@ -362,7 +362,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 						repository,
 						"<aggregate>",
 						"<instance-a>",
-						item0.Envelope.MetaData.MessageId,
+						"",
 					)
 					gomega.Expect(items).To(gomega.HaveLen(2))
 				}
--- a/persistence/internal/providertest/eventstore/repository.go
+++ b/persistence/internal/providertest/eventstore/repository.go
@@ -318,15 +318,11 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 					tc.Context,
 					"<aggregate>",
 					"<instance-a>",
-					item2.ID(),
+					"<unknown>",
 				)
 				gomega.Expect(err).Should(gomega.HaveOccurred())
 				gomega.Expect(err).To(
-					gomega.MatchError(
-						"message with ID '" +
-							item2.ID() +
-							"' cannot be found",
-					),
+					gomega.MatchError("message with ID '<unknown>' cannot be found"),
 				)
 			})
 
--- a/persistence/internal/providertest/eventstore/repository.go
+++ b/persistence/internal/providertest/eventstore/repository.go
@@ -306,6 +306,23 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 				}
 			})
 
+			ginkgo.It("returns an error if the message with the given id is not found", func() {
+				saveEvents(
+					tc.Context,
+					dataStore,
+					item0.Envelope,
+					item1.Envelope,
+				)
+
+				_, err := repository.LoadEventsBySource(
+					tc.Context,
+					"<aggregate>",
+					"<instance-a>",
+					item2.Envelope.MetaData.MessageId,
+				)
+				gomega.Expect(err).Should(gomega.HaveOccurred())
+			})
+
 			ginkgo.It("does not return an error if events exist beyond the end offset", func() {
 				res, err := repository.LoadEventsBySource(tc.Context, "<aggregate>", "<instance-a>", "")
 				gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
@@ -376,7 +393,6 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 				go fn2()
 				g.Wait()
 			})
-
 		})
 
 		ginkgo.Describe("type eventstore.Result", func() {
--- a/persistence/internal/providertest/eventstore/repository.go
+++ b/persistence/internal/providertest/eventstore/repository.go
@@ -278,7 +278,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 					repository,
 					"<aggregate>",
 					"<instance-a>",
-					item0.Envelope.MetaData.MessageId,
+					item0.ID(),
 				)
 
 				for i, item := range items {
@@ -294,7 +294,7 @@ func DeclareRepositoryTests(tc *common.TestContext) {
 					repository,
 					"<aggregate>",
 					"<instance-b>",
-					item1.Envelope.MetaData.MessageId,
+					item1.ID(),
 				)
 
 				for i, item := range items {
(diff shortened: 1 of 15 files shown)


jmalloc:
FYI I just changed the base branch for this PR from `master` to `220-integration`, it's pretty cool you can just edit it from the same place you edit the PR title.

persistence/subsystem/eventstore/repository.go
```diff
@@ -9,6 +9,16 @@ type Repository interface {
 	// NextEventOffset returns the next "unused" offset within the store.
 	NextEventOffset(ctx context.Context) (uint64, error)
 
+	// LoadEventsBySource loads the events produced by the specified source with
+	// the given handler key and id.
+	//
+	// d is the optional parameter, it represents ID of the message that was
+	// recorded when the instance was last destroyed.
```
jmalloc, line 16:
```suggestion
	// LoadEventsBySource loads the events produced by a specific handler.
	//
	// hk is the handler's identity key.
	//
	// id is the instance ID, which must be empty if the handler type does not
	// use instances.
	//
	// m is ID of a "barrier" message. If supplied, the results are limited to
	// events with higher offsets than the barrier message.
```

Since we decided to make this `LoadEventsBySource()`, rather than `LoadEventsForAggregate()`, I felt that the `d` parameter was too aggregate-specific, so I've tried to reword the docs just to "keep this layer dumb". Nothing I've written there changes the necessary behavior in any way.

persistence/provider/sql/sqlite/eventstore.go
```diff
@@ -289,6 +290,83 @@ func (driver) SelectEvents(
...
+			e.media_type,
+			e.data
+		FROM event AS e
+		WHERE e.source_app_key = $1
+		AND e.source_handler_key = $2
+		AND e.source_instance_id = $3
+		AND e.offset > (
+			SELECT e1.offset
+			FROM event AS e1
+			WHERE e1.message_id = $4
```
jmalloc, line 360: Looking at how this is implemented, I can't help but think we'd be better off having a driver method like `SelectOffsetByMessageID()`, calling that first to get the offset, then passing that offset to `SelectEventsBySource()`.

The downside is that there's an extra round trip, but it in practice it would only happen when the aggregate isn't cached or snapshotted.

The upside is it that it lets us reuse `SelectEventsBySource()` which is really the same query. More importantly, it gives us an opportunity to properly handle the failure condition that would occur if we can't actually find that message ID. It may also reduce the number of size of indices we need on the event table.

persistence/provider/sql/sqlite/schema.go
```diff
@@ -113,7 +113,7 @@ func createEventStoreSchema(ctx context.Context, db *sql.DB) {
 		`CREATE INDEX repository_query ON event (
 			source_app_key,
 			portable_name,
-			offset,
+			message_id,
```
jmalloc, line 116: Removing `offset` from this index means that `QueryEvents()` no longer has a good index to use.

There should be a separate index on the table for each variant of the query, and we can name them a little like the driver method they relate to. 

Please see [this comment](https://github.com/dogmatiq/infix/pull/253/files#diff-740d83e6a73bbe31f33fc03cfc0ab57bR360) before changing this, as it might change how many driver methods we have.

persistence/provider/sql/sqlite/eventstore.go
danilvpetrov: Yeah those subqueries felt like something out-of-line when I wrote them. The only goal I was pursuing at that time was to build autonomous methods enough to break away from the`eventstore.Query` usage and that dynamic query builder that is present in every implementation of `sql.Driver.QueryEvents`. `SelectOffsetByMessageID()` still does not break that tendency and that's so awesome.

Following the proposal, `SelectEventsBySource()` would now have to have an offset parameter. That might present a problem as this parameter would essentially mean all events with the offset greater than the given. When not used ( offset param == 0) this parameter will exclude the event with zero offset unless we code it as a dynamic query similar to  `sql.Driver.QueryEvents` implementations which we are trying to avoid.

I do admit that my concerns might be flawed 😄.

jmalloc: > When not used ( offset param == 0) this parameter will exclude the event with zero offset unless we code it as a dynamic query similar to `sql.Driver.QueryEvents` implementations which we are trying to avoid.

All we need to do is make the `WHERE` clause use `offset >= ?` and then add `1` to the offset of the "barrier" message, I'm pretty sure.

By the way, are you happy with that terminology - "barrier"?  I've seen it used to refer to offsets in the open/exclusive ranges we talked about with the `[x, y)` notation.

danilvpetrov: Yeah man barrier seems logical to refer to offset limit

danilvpetrov: It also sounds like we should return an error if the 'barier' message with the given ID is not found. I am also gonna add a test to address this case and change `boltdb` and `memory` to behave consistently.

Also, Can discuss the names of the SQL indices and their constituent fields when you have some free time?

jmalloc: > Can discuss the names of the SQL indices and their constituent fields when you have some free time?

Yep, just slack call whenever.

persistence/provider/sql/sqlite/schema.go
```diff
@@ -110,12 +110,21 @@ func createEventStoreSchema(ctx context.Context, db *sql.DB) {
 	sqlx.Exec(
 		ctx,
 		db,
-		`CREATE INDEX repository_query ON event (
+		`CREATE INDEX by_type ON event (
 			source_app_key,
 			portable_name,
-			offset,
+			"offset"
```
jmalloc, line 116: 🐼 

```suggestion
			offset
```

These don't need to be quoted in SQLite, for whatever reason. Just doing this to mimimize the diff/churn.

persistence/provider/sql/sqlite/schema.go
```diff
@@ -110,12 +110,21 @@ func createEventStoreSchema(ctx context.Context, db *sql.DB) {
...
+
+	sqlx.Exec(
+		ctx,
+		db,
+		`CREATE INDEX by_source ON event (
+			source_app_key,
 			source_handler_key,
-			source_instance_id
+			source_instance_id,
+			"offset"
```
jmalloc, line 127:
```suggestion
			offset
```

persistence/provider/sql/eventstore.go
```diff
@@ -64,20 +64,40 @@ type eventStoreDriver interface {
...
+	// ID.
+	SelectOffsetByMessageID(
+		ctx context.Context,
+		db *sql.DB,
+		id string,
+	) (uint64, error)
+
+	// ScanEvent scans the next event from a row-set returned by
+	// SelectEventsByType(), SelectEventsBySource(), and
+	// SelectEventsBySourceAfterMessage().
```
jmalloc, line 100:
```suggestion
	// ScanEvent scans the next event from a row-set returned by
	// SelectEventsByType() and SelectEventsBySource().
```

persistence/provider/sql/eventstore.go
```diff
@@ -64,20 +64,40 @@ type eventStoreDriver interface {
...
 		db *sql.DB,
 		ak string,
 		q eventstore.Query,
 		f int64,
 	) (*sql.Rows, error)
 
-	// ScanEvent scans the next event from a row-set returned by SelectEvents().
+	// SelectEventsBySource selects events from the eventstore that match the
+	// given source, namely the source's key and id. o is an optional offset
+	// parameter from which the messages should be loaded.
```
jmalloc, line 82:
```suggestion
	// SelectEventsBySource selects events from the eventstore that were
	// produced by a specific handler.
```

Trimmed this down, `o` isn't optional.

persistence/provider/sql/eventstore.go
```diff
@@ -64,20 +64,40 @@ type eventStoreDriver interface {
 		ak string,
 	) (uint64, error)
 
-	// SelectEvents selects events from the eventstore that match the given
-	// query.
+	// SelectEventsByType selects events from the eventstore that match the
+	// given event types query.
```
jmalloc, line 68:
```suggestion
	// given query.
```

Just changed this one back, since no behavior has changed there yet. This rename should probs have been a separate PR, but all good.

As usual, please duplicate these changes to the implementations in each provider.

persistence/provider/memory/eventstore.go
```diff
@@ -42,14 +42,53 @@ func (r *eventStoreRepository) NextEventOffset(
...
+// d is the optional parameter, it represents ID of the message that was
+// recorded when the instance was last destroyed.
+func (r *eventStoreRepository) LoadEventsBySource(
+	ctx context.Context,
+	hk, id, d string,
+) (eventstore.Result, error) {
+	var o uint64
+
+	if d != "" {
+		o1, ok := r.db.event.messageIDs[d]
```
jmalloc, line 57:
```suggestion
		var bool ok
		o, ok = r.db.event.messageIDs[d]
```

persistence/subsystem/eventstore/repository.go
```diff
@@ -9,6 +9,20 @@ type Repository interface {
...
+	// LoadEventsBySource loads the events produced by a specific handler.
+	//
+	// hk is the handler's identity key.
+	//
+	// id is the instance ID, which must be empty if the handler type does not
+	// use instances.
+	//
+	// m is ID of a "barrier" message. If supplied, the results are limited to
+	// events with higher offsets than the barrier message.
+	LoadEventsBySource(
```
jmalloc, line 21: Can you please propagate these docs to each of the implementations.

persistence/provider/memory/eventstore.go
```diff
@@ -98,7 +137,6 @@ func (r *eventStoreResult) Next(
 
 // Close closes the cursor.
 func (r *eventStoreResult) Close() error {
-	r.query.MinOffset = math.MaxUint64
```
jmalloc, line 101: You the equivalent line in the BoltDB implementation, any reason not to keep this one? Or otherwise, any reason not to get rid of both of them?

They're not strictly necessary, I just added them as a bit of protection to make sure you couldn't read anything after calling `Close()`, but I guess this one never really did anything.

persistence/provider/memory/eventstore.go
```diff
@@ -42,14 +42,53 @@ func (r *eventStoreRepository) NextEventOffset(
...
+			)
+		}
+
+		o = o1
+	}
+
+	return &eventStoreResult{
+		db: r.db,
+		pred: func(i *eventstore.Item) bool {
+			ok := hk == i.Envelope.MetaData.Source.Handler.Key &&
```
jmalloc, line 71:
```suggestion
			return hk == i.Envelope.MetaData.Source.Handler.Key &&
```

persistence/provider/memory/eventstore.go
jmalloc: 😱 We don't need no numbered variables 'round these parts!

persistence/provider/boltdb/eventstore.go
```diff
@@ -110,6 +129,66 @@ func (r *eventStoreRepository) NextEventOffset(
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return &eventStoreResult{
+		db:     r.db,
+		appKey: r.appKey,
+		pred: func(i *eventstore.Item) bool {
+			ok := hk == i.Envelope.MetaData.Source.Handler.Key &&
```
jmalloc, line 179:
```suggestion
			return hk == i.Envelope.MetaData.Source.Handler.Key &&
```

persistence/provider/boltdb/eventstore.go
```diff
@@ -110,6 +129,66 @@ func (r *eventStoreRepository) NextEventOffset(
...
+					if v == nil {
+						err = fmt.Errorf(
+							"message with id %s is not found",
+							d,
+						)
+
+						return
+					}
+
+					o = unmarshalUint64(v)
```
jmalloc, line 166:
```suggestion
					o = unmarshalUint64(v) + 1
```

persistence/provider/boltdb/eventstore.go
```diff
@@ -110,6 +129,66 @@ func (r *eventStoreRepository) NextEventOffset(
...
+
+	return &eventStoreResult{
+		db:     r.db,
+		appKey: r.appKey,
+		pred: func(i *eventstore.Item) bool {
+			ok := hk == i.Envelope.MetaData.Source.Handler.Key &&
+				id == i.Envelope.MetaData.Source.InstanceId
+
+			if d != "" {
+				return ok && i.Offset > o
```
jmalloc, line 183: This *shouldn't* be necessary at all, because `r.offset` should already take care of it. But I think this is currently "fixing" the fact that the offset isn't incremented +1 (as per the comment above). That's what it looks like anyway, I haven't tried changing any of it.

Same thing is going on in the in-memory provider.

persistence/provider/boltdb/eventstore.go
```diff
@@ -24,6 +25,13 @@ var (
 	// buffers.
 	eventStoreItemsBucketKey = []byte("items")
 
+	// eventStoreMesssageIDsBucketKey is the key for a child bucket that indexes
+	// event message id against the event offsets.
+	//
+	// The keys are the string message IDs converted to bytes. The
+	// values are the event offsets encoded as 8-byte big-endian packets.
+	eventStoreMesssageIDsBucketKey = []byte("message_ids")
```
jmalloc, line 33:
```suggestion
	eventStoreMessageIDsBucketKey = []byte("message_ids")
```

persistence/provider/boltdb/eventstore.go
```diff
@@ -24,6 +25,13 @@ var (
 	// buffers.
 	eventStoreItemsBucketKey = []byte("items")
 
+	// eventStoreMesssageIDsBucketKey is the key for a child bucket that indexes
```
jmalloc, line 28:
```suggestion
	// eventStoreMessageIDsBucketKey is the key for a child bucket that indexes
```

🐍

persistence/internal/providertest/eventstore/repository.go
```diff
@@ -196,30 +225,106 @@ func DeclareRepositoryTests(tc *common.TestContext) {
...
+					item1.Envelope,
+				)
+
+				_, err := repository.LoadEventsBySource(
+					tc.Context,
+					"<aggregate>",
+					"<instance-a>",
+					item2.Envelope.MetaData.MessageId,
+				)
+				gomega.Expect(err).Should(gomega.HaveOccurred())
```
jmalloc, line 323: I've tried to avoid checking for non-specific errors in the public API. If there's an error condition that the provider HAS to produce, I've opted to make a custom error type part of the public API. 

There are a few benefits:

1) It makes the tests easier to understand, just because it's more explicit.
2) I've had real cases where such tests were returning some error, but it was for a totally different reason, and I had no idea.
3) It makes it that little bit clearer for implementors of new persistence providers (you :P)

So, that said, we'll need an `eventstore.UnknownMessageError` type, and to add this to the docs for `LoadEventsBySource()`.

danilvpetrov: Cool, makes sense, will do.

persistence/provider/memory/eventstore.go
danilvpetrov: I am happy with either leaving or getting rid of them both.

jmalloc: Ok, if the tests pass without them, just get rid of them. I am happy to leave the behavior of a closed cursor's `Next()` method as undefined.

I'm working on the theory that if it had seemed important at the time I would have added a test for it.

danilvpetrov: Haha 😄 , just added that one back before I saw your message. Let me take them both out and re-run the tests.

danilvpetrov: Ok, local tests passed.

persistence/provider/memory/eventstore.go
```diff
@@ -125,18 +161,26 @@ func (cs *eventStoreChangeSet) stageSave(
...
 	cs.items = append(cs.items, item)
-
 	return next
 }
 
 // eventStoreDatabase contains data that is committed to the event store.
 type eventStoreDatabase struct {
-	items []*eventstore.Item
+	items      []*eventstore.Item
+	messageIDs map[string]uint64
```
jmalloc, line 170:
```suggestion
	offsets map[string]uint64
```

Makes more sense to name these after the value they store, IMO, just as `items` is.

persistence/provider/sql/eventstore.go
```diff
@@ -145,6 +163,52 @@ func (r *eventStoreRepository) NextEventOffset(
...
+			ctx,
+			r.db,
+			d,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		// increment the offset to select all messages after the barier 'd'
+		// message exclusively.
```
jmalloc, line 195:
```suggestion
		// Increment the offset to select all messages after the barrier
		// message exclusively.
```

persistence/provider/memory/eventstore.go
```diff
@@ -42,14 +41,52 @@ func (r *eventStoreRepository) NextEventOffset(
...
+	if d != "" {
+		var ok bool
+		o, ok = r.db.event.messageIDs[d]
+		if !ok {
+			return nil, eventstore.UnknownMessageError{
+				MessageID: d,
+			}
+		}
+
+		o = o + 1
```
jmalloc, line 69: 🐧 
```suggestion
		// Increment the offset to select all messages after the barrier
		// message exclusively.
		o++
```

Just to make this match the SQL impl - you had the comment explaining why there, and I thought that was good.

persistence/provider/memory/eventstore.go
```diff
@@ -42,14 +41,52 @@ func (r *eventStoreRepository) NextEventOffset(
...
+//
+// id is the instance ID, which must be empty if the handler type does not
+// use instances.
+//
+// m is ID of a "barrier" message. If supplied, the results are limited to
+// events with higher offsets than the barrier message. If the message
+// cannot be found, UnknownMessageError is returned.
+func (r *eventStoreRepository) LoadEventsBySource(
+	ctx context.Context,
+	hk, id, d string,
```
jmalloc, line 56:
```suggestion
	hk, id, m string,
```

persistence/provider/boltdb/eventstore.go
```diff
@@ -110,6 +127,64 @@ func (r *eventStoreRepository) NextEventOffset(
...
+					v := messageIDs.Get([]byte(d))
+					if v == nil {
+						err = &eventstore.UnknownMessageError{
+							MessageID: d,
+						}
+
+						return
+					}
+
+					o = unmarshalUint64(v) + 1
```
jmalloc, line 168: 🐧 

```suggestion
					// Increment the offset to match all messages
					// after the barrier message exclusively.
					o = unmarshalUint64(v) + 1
```

Just to make this match the SQL impl - you had the comment explaining why there, and I thought that was good.

persistence/provider/boltdb/eventstore.go
```diff
@@ -110,6 +127,64 @@ func (r *eventStoreRepository) NextEventOffset(
...
+//
+// id is the instance ID, which must be empty if the handler type does not
+// use instances.
+//
+// m is ID of a "barrier" message. If supplied, the results are limited to
+// events with higher offsets than the barrier message. If the message
+// cannot be found, UnknownMessageError is returned.
+func (r *eventStoreRepository) LoadEventsBySource(
+	ctx context.Context,
+	hk, id, d string,
```
jmalloc, line 142:
```suggestion
	hk, id, m string,
```

persistence/provider/boltdb/eventstore.go
```diff
@@ -24,6 +23,13 @@ var (
 	// buffers.
 	eventStoreItemsBucketKey = []byte("items")
 
+	// eventStoreMessageIDsBucketKey is the key for a child bucket that indexes
+	// event message id against the event offsets.
+	//
+	// The keys are the string message IDs converted to bytes. The
+	// values are the event offsets encoded as 8-byte big-endian packets.
+	eventStoreMessageIDsBucketKey = []byte("message_ids")
```
jmalloc, line 31:
```suggestion
	eventStoreOffsetsBucketKey = []byte("offsets")
```

I made a similar suggestion on the memory implementation - we typically name the buckets for what they contain, now how they're keyed, if that makes sense.

persistence/internal/providertest/eventstore/repository.go
```diff
@@ -196,30 +225,113 @@ func DeclareRepositoryTests(tc *common.TestContext) {
...
+				for i, item := range items {
+					expectItemToEqual(
+						item,
+						expectedB[i],
+						fmt.Sprintf("item at index #%d of slice", i),
+					)
+				}
+			})
+
+			ginkgo.It("returns an error if the message with the given id is not found", func() {
```
jmalloc, line 309:
```suggestion
			ginkgo.It("returns an error if the barrier message is not found", func() {
```

🐧

persistence/internal/providertest/eventstore/repository.go
```diff
@@ -196,30 +225,113 @@ func DeclareRepositoryTests(tc *common.TestContext) {
...
+					)
 				}
+			})
 
-				g.Add(3)
-				go fn()
-				go fn()
-				go fn()
-				g.Wait()
+			ginkgo.It("matches the events from the source only after the given message id", func() {
```
jmalloc, line 261:
```suggestion
			ginkgo.It("only matches the events after the barrier message", func() {
```

🐧

persistence/provider/sql/eventstore.go
```diff
@@ -64,20 +64,38 @@ type eventStoreDriver interface {
...
+		o uint64,
+	) (*sql.Rows, error)
+
+	// SelectOffsetByMessageID selects the offset of the message with the given
+	// ID. It returns eventstore.UnknownMessageError if the message is not found.
+	SelectOffsetByMessageID(
+		ctx context.Context,
+		db *sql.DB,
+		id string,
+	) (uint64, error)
```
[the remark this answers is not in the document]
danilvpetrov: Yeah that's very distinct way of doing that. The only problem I see with this approach is the coverage of the error check in the code.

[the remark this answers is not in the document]
jmalloc: How do you mean?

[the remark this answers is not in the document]
danilvpetrov: Ok, this one should be good to go.

persistence/provider/boltdb/eventstore.go
```diff
@@ -68,7 +68,7 @@ func (t *transaction) SaveEvent(
 
 	messageIDs := bboltx.CreateBucketIfNotExists(
```
jmalloc, line 69:
```suggestion
	offsets := bboltx.CreateBucketIfNotExists(
```

persistence/provider/boltdb/eventstore.go
```diff
@@ -139,32 +139,34 @@ func (r *eventStoreRepository) NextEventOffset(
...
 		o   uint64
 		err error
 	)
 
-	if d != "" {
+	if m != "" {
 		r.db.View(
 			ctx,
 			func(tx *bbolt.Tx) {
 				if messageIDs, exists := bboltx.TryBucket(
```
jmalloc, line 153:
```suggestion
				if offsets, exists := bboltx.TryBucket(
```

persistence/provider/boltdb/eventstore.go
```diff
@@ -60,6 +66,17 @@ func (t *transaction) SaveEvent(
 
 	saveEventStoreItem(items, o, env)
 
+	messageIDs := bboltx.CreateBucketIfNotExists(
+		store,
+		eventStoreOffsetsBucketKey,
+	)
+
+	bboltx.Put(
+		messageIDs,
```
jmalloc, line 75:
```suggestion
		offsets,
```

persistence/provider/boltdb/eventstore.go
```diff
@@ -110,6 +127,66 @@ func (r *eventStoreRepository) NextEventOffset(
...
+		r.db.View(
+			ctx,
+			func(tx *bbolt.Tx) {
+				if messageIDs, exists := bboltx.TryBucket(
+					tx,
+					r.appKey,
+					eventStoreBucketKey,
+					eventStoreOffsetsBucketKey,
+				); exists {
+					v := messageIDs.Get([]byte(m))
```
jmalloc, line 159:
```suggestion
					v := offsets.Get([]byte(m))
```

persistence/provider/memory/eventstore.go
```diff
@@ -125,18 +163,26 @@ func (cs *eventStoreChangeSet) stageSave(
...
 // apply updates the database to include the changes in cs.
 func (db *eventStoreDatabase) apply(cs *eventStoreChangeSet) {
 	db.items = append(db.items, cs.items...)
+
+	if db.offsets == nil {
+		db.offsets = make(map[string]uint64)
+	}
+
+	for _, item := range cs.items {
+		db.offsets[item.Envelope.MetaData.MessageId] = item.Offset
```
jmalloc, line 184:
```suggestion
		db.offsets[item.ID()] = item.Offset
```

I put these `ID()` convenience methods on the items.  I should probs add one to `Envelope` and `Parcel` as well.

persistence/internal/providertest/eventstore/repository.go
```diff
@@ -196,30 +225,113 @@ func DeclareRepositoryTests(tc *common.TestContext) {
...
+					item2.ID(),
+				)
+				gomega.Expect(err).Should(gomega.HaveOccurred())
+				gomega.Expect(err).To(
+					gomega.MatchError(
+						"message with ID '" +
+							item2.ID() +
+							"' cannot be found",
+					),
+				)
```
jmalloc, line 330:
```suggestion
					"<unknown>",
				)
				gomega.Expect(err).Should(gomega.HaveOccurred())
				gomega.Expect(err).To(
					gomega.MatchError("message with ID '<unknown>' cannot be found"),
				)
```

PR #258, opened 2020-05-09
Master PR for persistence refactor.

### LINKED ISSUE #123: Change internal persistence tests to use the TestContext.
The signature of these functions in `persistence/internal/providertest` need to match the signatures of the `DeclareXXXTests()` functions in the subpackages (`eventstore`, for example).

- `declareProviderTests()`
- `declareDataStoreTests()`
- `declareTransactionTests()`

### LINKED ISSUE #151: Replace custom expects in provider tests with EqualX().

### LINKED ISSUE #168: Remove offset return value from `SaveEvent()`.

### LINKED ISSUE #181: Clean up BoltDB internals.
The different variants of various marshaling, loading and saving functions are getting difficult to understand.

### LINKED ISSUE #202: Save events directly to the queue if they are consumed within the same app.

### LINKED ISSUE #220: Improve relationship between aggregate instances and their events.

### LINKED ISSUE #222: Defer insertion of events until Commit() in SQL providers.
Blocked by #220

### LINKED ISSUE #241: Risk of future race condition in aggregate.Sink.
When `aggregate.Sink` persists aggregate meta-data it increments the revision immediately after the call to `SaveAggregateMetaData()`. This is done so that the revision is correct when the instance is returned to the cache. However, this approach does not directly take into consideration what happens when **(a)** the transaction is rolled back, **(b)** another goroutine handles a message for the same instance *after* the instance is returned to the cache, but *before* the transaction is committed.

As it turns out, these are both non-issues right now, but this is more a matter of luck than good planning:

**(a) The transaction is rolled back** - the only operations performed on the transaction after the meta-data update are removing the causal message from the queue, and committing the transaction. Both of these operations are performed inside `queue.Request.Ack()`, which is in turn called by the `pipeline.Ack()` stage. As it stands, if `Ack()` returns an error the entire engine is stopped altogether, obliterating the cache.

**(b) Another goroutine handles a message for the same instance** - in this scenario, we get away with the current implementation only because all of the current persistence providers implementations use some kind of row or in-memory locking, guaranteeing that `ErrConflict` will be returned by `SaveAggregateMetaData()`, and not by `Commit()`.

  - goroutine A updates meta-data to revision 123
  - goroutine A releases cache lock
  - goroutine B acquires cache lock
  - goroutine B updates meta-data to revision 124 (this blocks)
  - goroutine A commits the transaction 
  - goroutine B unblocks and continues

---

I don't like either of these situations because they aren't covered by any tests. (b) also blocks #175, because with DynamoDB the entire transaction will need to be submitted as a batch on commit, meaning that `SaveAggregateMetaData()` will have no opportunity to lock any "row" or return `ErrConflict`.

### LINKED ISSUE #242: Disallow manipulation each "entity" more than once within a single transaction.
Blocks #175

In order to allow this behavior (which we don't need) and still be able to implement providers like DynamoDB (which requires transactions to be submitted in a single HTTP request), we'd have to keep track of the revisions of each "entity" (queue item, aggregate meta-data, etc) in memory while the transaction is in-flight (a bit like the memory provider does now). We'd also have to allow `ErrConflict` errors to be returned from both the individual save operations, and `Commit()`.

Instead, we can disallow saving/updating/deleting the same entity within the same transaction by panicking if this is attempted. There's no valid use case where this can happen, and it's managed by the engine, not the Dogma application code. There'd have to be a UUID collision or something like that for it to occur in an otherwise bug-free engine.

We'd still have to maintain a list of which entities have been modified already but that's a lot simpler than tracking the revisions.

### LINKED ISSUE #251: Refactor transaction to be represented as a batch of operation structs.

jmalloc:
- Fixes #123 - Change internal persistence tests to use the TestContext.
- Fixes #151 - Replace custom expects in provider tests with EqualX().
- Fixes #168 - Remove offset return value from `SaveEvent()`
- Fixes #181 - Clean up BoltDB internals
- Fixes #202 - Save events directly to the queue if they are consumed within the same app
- Fixes #220 - Improve relationship between aggregate instances and their events
- Fixes #222 - Defer insertion of events until Commit() in SQL provider
- Fixes #241 - Risk of future race condition in aggregate.Sink.
- Fixes #251 - Refactor transaction to be represented as a batch of operations
- Fixes #242 - Disallow multiple operations for the same "entity", see `Batch.MustValidate()


jmalloc:
FYI @danilvpetrov I've rebased this integration branch off `master` and will do so periodically. This should have much impact so long as we keep working in PRs off this branch, rather than modifying this branch directly.

PR #262, opened 2020-05-11
Add "batch support" to the pipeline package.

jmalloc:
This PR is another stepping-stone towards #251. It modifies the pipeline package to allow for use of persistence batches alongside conventional transactions.


jmalloc:
I'm not too worried about the coverage on these intermediate ones.

I'm still adding tests for new code that I expect to stick around, but there's a few uncovered bits that will be refactored away before `220-integration` is merged to `master`.

PR #276, opened 2020-05-20
Collapse persistence subsystems into a single package.

### LINKED ISSUE #229: Replace `eventstore.Repository.QueryEvents()` with `LoadEventsByType()`.
Blocked by #220 

Once #220 is complete the only remaining fields in query would be the minimum offset and the type filter. `QueryEvents()` can be replaced with `LoadEventsByType(ctx, filter, offset)`. This method signature is analogous to that of `eventstream.Open()`.

jmalloc:
This PR collapses the packages under `persistence/subsystem` into the `persistence` package.

I've also fixed #229 at the same time, as it reduced the complexity of the event repository significantly.


jmalloc:
graph

PR #343, opened 2020-11-02
Automatically create aggregate instances when first event is recorded.

jmalloc:
This PR updates infix to work with the changes introduced to Dogma in https://github.com/dogmatiq/dogma/pull/129 (and tagged in Dogma v0.8.0).


danilvpetrov (approved):
That looks pretty straightforward.

koden-km (approved):

PR #345, opened 2020-11-05
Rename `AggregateMetaData.LastDestroyedBy` to `BarrierEventID`.

jmalloc:
This PR is in preparation for the changes documented in [ADR-17](https://github.com/dogmatiq/dogma/blob/master/docs/adr/0017-recreate-aggregate-after-destruction.md).

Because an aggregate instance can now be destroyed without recording a new event, the `AggregateMetaData.LastDestroyedByID`, which is used to work out the earliest historical event to load, is no longer necessarily the event that *caused* the destruction, it's just the last event *before* destruction.

The persistence layer refers to this message as a "barrier message" elsewhere, so I am using that terminology here too.


jmalloc:
I wonder if this rings any bells now, Danil :)

danilvpetrov (approved):
It actually does 😁. I do remember the aggregate schema!

PR #357, opened 2020-11-17
Compact projections periodically.

### LINKED ISSUE #354: Implement projection compaction.
Dogma v0.10.0 introduced the notion of projection compaction.

We need to develop a scheme for when compaction should be performed. I would like to keep it fairly simple. 

We could just do exactly what Aperture's [projector](https://github.com/dogmatiq/aperture/blob/master/ordered/projector.go#L68) does. It runs compaction on a fixed interval with a timeout. However, in a more substantial application with many projections this might lead to a lot of compaction happening at more or less random intervals. 

Another approach would be to run compaction after events have been handled with some kind of debounce period.  This is a bit smarter but hopefully still fairly easy to implement and understand.

For example, after an event has been handled we could start a timer. Whenever a new event is received the timer is reset. If the timer fires without being reset we perform compaction. We'd need some upper cap on how long to go between compactions so that it doesn't get deferred forever, and so it is performed even when no events are being handled.

jmalloc:

This PR implements projection compaction, as introduced in Dogma v0.10.0.

In the interest of simplicity I ended up just running compaction at a fixed interval. I had made some more advanced proposals in #354.

Every projection is compacted on start up. To ensure that applications with a large number of projections do not stampede the database the "compactors" honour the same concurrency limit as message handlers (see the `WithConcurrencyLimit()` engine option). 

Fixes #354

#### Why make this change?

Compaction was introduced to clean-up application read-model data. It is the engine's responsibility to schedule this compaction.

I've used a rather generous compaction timeout of 5 minutes. This is the default in `dogmatiq/aperture` as well. This should perhaps be reduced but it's hard to say. The Dogma documentation suggests that compaction be performed as multiple small transactions so that large projections can at least be "partially compacted" within the timeout period.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1,0 +1,47 @@
+package verity
+
+import (
+	"context"
+
+	"github.com/dogmatiq/configkit"
+	"github.com/dogmatiq/verity/handler/projection"
+	"github.com/dogmatiq/verity/internal/x/loggingx"
+	"golang.org/x/sync/errgroup"
+)
+
+// runCompactorsForApp runs a projection compactor for a each projection with
+// a specific application.
+func (e *Engine) runCompactorsForApp(ctx context.Context, a *app) error {
+	g, ctx := errgroup.WithContext(ctx)
+
+	for _, h := range a.Config.RichHandlers().Projections() {
+		h := h // capture loop variable
+		g.Go(func() error {
+			return e.runCompactorForProjection(ctx, h, a)
+		})
+	}
+
+	return g.Wait()
+}
+
+// runCompactorForProjection runs a projection compactor for a specific
+// projection.
+func (e *Engine) runCompactorForProjection(
+	ctx context.Context,
+	h configkit.RichProjection,
+	a *app,
+) error {
+	c := &projection.Compactor{
+		Handler:  h.Handler(),
+		Interval: e.opts.ProjectionCompactInterval,
+		Timeout:  e.opts.ProjectionCompactTimeout,
+		Logger: loggingx.WithPrefix(
+			e.logger,
+			"[compactor %s@%s] ",
+			h.Identity().Name,
+			a.Config.Identity().Name,
+		),
+	}
+
+	return c.Run(ctx)
+}
--- a/compaction.go
+++ b/compaction.go
@@ -32,9 +32,10 @@ func (e *Engine) runCompactorForProjection(
 	a *app,
 ) error {
 	c := &projection.Compactor{
-		Handler:  h.Handler(),
-		Interval: e.opts.ProjectionCompactInterval,
-		Timeout:  e.opts.ProjectionCompactTimeout,
+		Handler:   h.Handler(),
+		Interval:  e.opts.ProjectionCompactInterval,
+		Timeout:   e.opts.ProjectionCompactTimeout,
+		Semaphore: e.semaphore,
 		Logger: loggingx.WithPrefix(
 			e.logger,
 			"[compactor %s@%s] ",
--- a/compaction.go
+++ b/compaction.go
@@ -9,7 +9,7 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-// runCompactorsForApp runs a compactor for a each projection within a specific
+// runCompactorsForApp runs a compactor for each projection within a specific
 // application.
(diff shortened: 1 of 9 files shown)


engineoption.go
```diff
@@ -44,11 +44,23 @@ var (
...
-	// concurrently.
+	// (and projections to compact) concurrently.
 	//
 	// It is overridden by the WithConcurrencyLimit() option.
 	DefaultConcurrencyLimit = uint(runtime.GOMAXPROCS(0) * 2)
 
+	// DefaultProjectionCompactInterval is the default interval at which the
+	// engine compacts projections.
+	//
+	// It is overridden by the WithProjectionComapctInterval() option.
```
danilvpetrov:
```suggestion
	// It is overridden by the WithProjectionCompactInterval() option.
```

engineoption.go
```diff
@@ -44,11 +44,23 @@ var (
...
+	DefaultProjectionCompactInterval = 24 * time.Hour
+
+	// DefaultProjectionCompactTimeout is the default timeout to use when
+	// compacting a projection.
+	//
+	// It is overridden by the WithProjectionComapctTimeout() option.
```
danilvpetrov:
```suggestion
	// It is overridden by the WithProjectionCompactTimeout() option.
```

danilvpetrov (approved):
Few spelling corrections, everything else looks good.

compaction.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"context"
+
+	"github.com/dogmatiq/configkit"
+	"github.com/dogmatiq/verity/handler/projection"
+	"github.com/dogmatiq/verity/internal/x/loggingx"
+	"golang.org/x/sync/errgroup"
+)
+
+// runCompactorsForApp runs a compactor for a each projection within a specific
```
koden-km:
```suggestion
// runCompactorsForApp runs a compactor for each projection within a specific
```

engineoption.go
```diff
@@ -130,6 +143,36 @@ func WithConcurrencyLimit(n uint) EngineOption {
 	}
 }
 
+// WithProjectionCompactInterval returns an engine option that set the interval
```
koden-km:
```suggestion
// WithProjectionCompactInterval returns an engine option that sets the interval
```

engineoption.go
```diff
@@ -130,6 +143,36 @@ func WithConcurrencyLimit(n uint) EngineOption {
...
+		panic("duration must not be negative")
+	}
+
+	return func(opts *engineOptions) {
+		opts.ProjectionCompactInterval = d
+	}
+}
+
+// WithProjectionCompactTimeout returns an engine option that set the duration
+// the engine allows for a single projectiont to be compacted.
```
koden-km:
```suggestion
// WithProjectionCompactTimeout returns an engine option that sets the duration
// the engine allows for a single projection to be compacted.
```

handler/projection/compactor_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+				called = true
+				return nil
+			}
+
+			err := compactor.Run(ctx)
+			Expect(err).To(Equal(context.Canceled))
+		})
+
+		It("returns if the context is canceled while waiting for the sempahore", func() {
```
koden-km:
```suggestion
		It("returns if the context is canceled while waiting for the semaphore", func() {
```

Also, does this return an error or just returns?

koden-km (approved):
Just some doc blocks. Approved once fixed.

handler/projection/compactor_test.go
jmalloc: It returns the `context.Canceled` error.

handler/projection/compactor_test.go
```diff
@@ -0,0 +1,120 @@
...
+				dogma.ProjectionCompactScope,
+			) error {
+				return errors.New("<error>")
+			}
+
+			err := compactor.Run(context.Background())
+			Expect(err).To(MatchError("<error>"))
+		})
+
+		It("does not return an error compaction times out", func() {
```
jmalloc:
```suggestion
		It("does not return an error if compaction times out", func() {
```

handler/projection/compactor_test.go
jmalloc:
```suggestion
		It("returns an error if the context is canceled while waiting for the semaphore", func() {
```

PR #376, opened 2021-01-28
[WIP] Add compatibility for MariaDB 10.2

neetle:
From local testing, the proposed change allows verity to start up with both MariaDB 10.2 and 10.4. I'll need to verify the rest.


neetle:
Running the tests locally against `mariadb/server:10.2`, everything seems to work.

jmalloc:
Thanks man :+1:

I'm not sure why the tests haven't run here but I'll merge this anyway so you can move on.

PR #380, opened 2021-02-03
Implement event stream content negotiation.

jmalloc:

This PR adds support for basic content-type negotiation for event messages based on the MIME media-type.

The rules are:

1. If the client supports the existing encoding of the message, it is sent unchanged
2. Otherwise, it is re-encoded using the client's most-preferred media-type that is also supported by the server.

#### Why make this change?

This is somewhat of a future proofing measure that lets us consume events in different formats to how they are persisted in the event stream if necessary.

No

None

### DIFF
--- a/eventstream/networkstream/server.go
+++ b/eventstream/networkstream/server.go
@@ -4,10 +4,12 @@ import (
 	"context"
 
 	"github.com/dogmatiq/configkit/message"
+	"github.com/dogmatiq/interopspec/envelopespec"
 	"github.com/dogmatiq/interopspec/eventstreamspec"
 	"github.com/dogmatiq/marshalkit"
 	"github.com/dogmatiq/verity/eventstream"
 	"github.com/dogmatiq/verity/internal/x/grpcx"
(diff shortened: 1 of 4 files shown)


danilvpetrov (approved):
Nice and concise PR, I like your `github.com/dogmatiq/verity/internal/x/grpcx` package methods.

eventstream/networkstream/server.go
```diff
@@ -176,40 +183,97 @@ func (s *server) EventTypes(
...
+			failures,
+			"one or more unrecognized event types or media-types",
 		)
 	}
 
-	if len(types) == 0 {
+	return messageTypes, mediaTypes, nil
+}
+
+// transcope returns an envelope containing the message in p, transcoded into a
```
koden-km:
```suggestion
// transcode returns an envelope containing the message in p, transcoded into a
```

koden-km (approved):
Approved after the suggested typo fix.

jmalloc:
Thanks!

PR #390, opened 2021-03-11
Add `queue.StreamAdaptor`.

jmalloc:

This PR introduces `queue.StreamAdaptor`, which is an `eventstream.Handler` that places messages into a `queue.Queue`. It also includes the necessary glue code to consume events from each discovered application.

#### Why make this change?

This mechanism is required to obtain events produced by "other" applications that are consumed by process message handlers within "this" application.

No

None


jmalloc:
GitHub actions has an outage at the moment, I am merging this without checks.

PR #435, opened 2021-12-10
Implement aggregate snapshots

Evildethow:

Implement aggregate snapshots.

#### Why make this change?

Performance optimization.

TBD.

- https://github.com/dogmatiq/verity/issues/142


persistence/aggregate.go
```diff
@@ -32,6 +35,21 @@ type AggregateMetaData struct {
...
+
+	// HandlerKey is the identity key of the aggregate message handler.
+	HandlerKey string
+
+	// InstanceID is the aggregate instance ID.
+	InstanceID string
+
+	// Revision is the instance's current version, used to enforce optimistic
+	// concurrency control.
+	Revision uint64
```
jmalloc:
```suggestion
	// Revision is the instance's current version.
	Revision uint64
```

We don't need to enforce OCC when persisting a snapshot because it doesn't matter if it's out of date - we'll just replay any events that happened since on top of it. This has some positive implications, like the fact that we don't _have_ to persist it in the same transaction as other aggregate options.

persistence/aggregate.go
```diff
@@ -41,6 +59,14 @@ type AggregateRepository interface {
...
 		hk, id string,
 	) (AggregateMetaData, error)
+
+	// LoadAggregateSnapshot loads the latest snapshot for an aggregate instance.
+	//
+	// hk is the aggregate handler's identity key, id is the instance ID.
+	LoadAggregateSnapshot(
+		ctx context.Context,
+		hk, id string,
+	) (*AggregateSnapshot, error)
```
jmalloc: I'd avoid using a pointer here, because:
- we don't really want reference semantics, `AggregateSnapshot` is just a plain-old-value
- It's less clear to the caller that `nil` is meaningful (vs the `ok bool` style)
- marginally less GC pressure without the heap allocation

```suggestion
	) (AggregateSnapshot, bool, error)
```

persistence/aggregate.go
```diff
@@ -62,3 +88,23 @@ func (op SaveAggregateMetaData) AcceptVisitor(ctx context.Context, v OperationVi
...
 }
+
+// SaveAggregateSnapshot is an Operation that creates or updates snapshots of
+// an aggregate instance.
+type SaveAggregateSnapshot struct {
+	// Snapshot is the snapshot to persist.
+	//
+	// Snapshot.Revision must be the revision of the aggregate instance as
+	// currently persisted, otherwise an optimistic concurrency conflict occurs
+	// and the entire batch of operations is rejected.
```
jmalloc:
```suggestion
```

persistence/aggregate.go
```diff
@@ -32,6 +35,21 @@ type AggregateMetaData struct {
 	BarrierEventID string
 }
 
+type AggregateSnapshot struct {
+	// State is the instance's current aggregate state.
+	State dogma.AggregateRoot
```
jmalloc: > I'm a little unsure about how to best capture the current state for the snapshot. AFAIK, dogma.AggregateRoot needs to be persisted, but this may be incorrect.

Yep, this is correct. 

The [`persistence.ProcessInstance`](https://github.com/dogmatiq/verity/blob/main/persistence/process.go#L10) type might serve as a decent reference here. It's used to persist `dogma.ProcessRoot` instances, which is a pretty similar idea.

Note the use of [`marshalkit.Packet`](https://pkg.go.dev/github.com/dogmatiq/marshalkit#Packet). All user-defined data types (messages, aggregate roots, process roots, etc) are persisted using this structure. The upshot is that the `verity/persistence` package remains unaware of the user-defined types, it just operates on binary representations.

```suggestion
	// Packet contains the binary representation of the aggregate snapshot.
	Packet marshalkit.Packet
```

(Pedantic: for consistency with the other data types, I'd also move this field to the end of the struct).

jmalloc: See also [`verity.NewDefaultMarshaler()`](https://github.com/dogmatiq/verity/blob/b3ddea0ef27b300172bf33911ec64b807b20c4e7/engineoption.go#L181) which constructs a [`marshalkit.Marshaler`](https://pkg.go.dev/github.com/dogmatiq/marshalkit#Marshaler) to use for a given `dogma.Application` instance.  This will need to be modified to make the marshaler aware of the aggregate root types, again similar to the way the process roots work.

jmalloc (changes requested):

persistence/aggregate.go
Evildethow: @jmalloc 

RE:

> See also verity.NewDefaultMarshaler() which constructs a marshalkit.Marshaler to use for a given dogma.Application instance. This will need to be modified to make the marshaler aware of the aggregate root types, again similar to the way the process roots work.

Could you elaborate on this?

Are you suggesting that a new Marshaler needs to be created to accept unmarshaling of `dogma.AggregateRoot` type?

IIUC, `dogma.ProcessRoot` can be unmarshalled due to its type being a marker `interface{}` with no methods, where as `dogma.AggregateRoot` defines an `ApplyEvent` method, which results in a compile error when attempting to unmarshal to this type using the same process. 

See https://github.com/dogmatiq/verity/blob/main/handler/process/loader.go#L53

jmalloc: > Are you suggesting that a new Marshaler needs to be created to accept unmarshaling of dogma.AggregateRoot type?

Not a new marshaler no, I just mean that the `NewDefaultMarshaler()` function in Verity has to be modified to make the marshaler it constructs aware of the aggregate root types.

It's already aware of the process root types because of this code here: https://github.com/dogmatiq/verity/blob/b3ddea0ef27b300172bf33911ec64b807b20c4e7/engineoption.go#L188

handler/aggregate/adaptor.go
```diff
@@ -16,6 +17,8 @@ import (
 	"github.com/dogmatiq/verity/persistence"
 )
 
+const snapshotInterval = 1000
```
jmalloc, line 21:
```suggestion
// snapshotInterval is the number of revisions that may occur before saving a new snapshot.
const snapshotInterval = 1000
```

handler/aggregate/adaptor_test.go
```diff
@@ -248,6 +251,96 @@ var _ = Describe("type Adaptor", func() {
...
+					_ dogma.Message,
+				) {
+					s.RecordEvent(MessageE1)
+
+					r := x.(*AggregateRoot)
+					Expect(r.AppliedEvents).To(Equal(
+						[]dogma.Message{
+							MessageE1,
+						},
+					))
```
jmalloc: I _think_ this assertion was probs copied from one of the other tests, not directly related to this test.

```suggestion
```

handler/aggregate/adaptor_test.go
```diff
@@ -248,6 +251,96 @@ var _ = Describe("type Adaptor", func() {
...
```
jmalloc:
```suggestion
```

handler/aggregate/loader.go
```diff
@@ -42,6 +41,21 @@ func (l *Loader) Load(
...
+	}
+
+	if ok {
+		s, err := l.Marshaler.Unmarshal(ss.Packet)
+		if err != nil {
+			return nil, err
+		}
+
+		base = s.(dogma.AggregateRoot)
+		md.BarrierEventID = ss.Version
```
jmalloc: There's a subtle issue with setting `md.BarrierEventID` here. Doing so makes `l.applyEvents()` behave as desired but it will also be persisted in the `AggregateMetaData` on the next save which essentially marks the instance as destroyed as of that event.

This probably has no observable effect so long as the snapshot always remains available, but there's one particular case where snapshots may need to be removed, which is when the `AggregateRoot` implementation is changed to the point where old snapshots are no longer valid representation.

The "standard" approach in this scenario is to just delete the incompatible snapshots so that new snapshots are built from the event history, but because `BarrierEventID` has been set these new snapshots will not "see" the entire event history.

I would suggest adding an additional `barrierEventID` parameter to `applyEvents()`, pass it `md.BarrierEventID` by default, but override it with `ss.Version` if available, and avoid modifying `md.BarrierEventID` altogether.

handler/aggregate/loader_test.go
```diff
@@ -76,6 +90,12 @@ var _ = Describe("type Loader", func() {
 				))
 			})
 
+			It("does not return a snapshot", func() {
+				inst, err := loader.Load(ctx, "<handler-key>", "<instance>", base)
+				Expect(err).ShouldNot(HaveOccurred())
+				Expect(inst.BarrierEventID).To(BeEmpty()) // is set when snapshot is present
```
jmalloc: This line is what clued me into that issue with `md.BarrierEventID` I mentioned above, so I suspect this test will no longer function after that is fixed.

handler/aggregate/loader_test.go
```diff
@@ -130,6 +150,57 @@ var _ = Describe("type Loader", func() {
...
+					AppliedEvents: []dogma.Message{
+						MessageE1,
+						MessageE2,
+						MessageE3,
+					},
+				}
+
+				var packet marshalkit.Packet
+				var err error
+				packet, err = Marshaler.Marshal(ssBase)
```
jmalloc:
```suggestion
				packet, err := Marshaler.Marshal(ssBase)
```

Pedantic Go developer shit.

persistence/aggregate.go
```diff
@@ -32,6 +35,20 @@ type AggregateMetaData struct {
...
 
+type AggregateSnapshot struct {
+	// HandlerKey is the identity key of the aggregate message handler.
+	HandlerKey string
+
+	// InstanceID is the aggregate instance ID.
+	InstanceID string
+
+	// Version is the instance's current version.
+	Version string
```
jmalloc:
```suggestion
	// LastEventID is the most recent event ID applied to the aggregate root at the point this snapshot was taken.
	LastEventID string
```

Suggestion for clarity.

persistence/aggregate.go
```diff
@@ -62,3 +87,37 @@ func (op SaveAggregateMetaData) AcceptVisitor(ctx context.Context, v OperationVi
...
+	Snapshot AggregateSnapshot
+}
+
+// AcceptVisitor calls v.VisitSaveAggregateSnapshot().
+func (op SaveAggregateSnapshot) AcceptVisitor(ctx context.Context, v OperationVisitor) error {
+	return v.VisitSaveAggregateSnapshot(ctx, op)
+}
+
+func (op SaveAggregateSnapshot) entityKey() entityKey {
+	return entityKey{"handler", op.Snapshot.HandlerKey, op.Snapshot.InstanceID}
```
jmalloc:
```suggestion
	return entityKey{"snapshot", op.Snapshot.HandlerKey, op.Snapshot.InstanceID}
```

These operations will need "entity keys" distinct from `SaveAggregateMetaData`, otherwise they'll conflict and cause a panic when performed in the same `UnitOfWork`.

I suspect this simply isn't being caught because there's no end-to-end test of handler behaviour that uses real persistence providers.

persistence/aggregate.go
```diff
@@ -62,3 +87,37 @@ func (op SaveAggregateMetaData) AcceptVisitor(ctx context.Context, v OperationVi
...
+	Snapshot AggregateSnapshot
+}
+
+// AcceptVisitor calls v.VisitRemoveAggregateSnapshot().
+func (op RemoveAggregateSnapshot) AcceptVisitor(ctx context.Context, v OperationVisitor) error {
+	return v.VisitRemoveAggregateSnapshot(ctx, op)
+}
+
+func (op RemoveAggregateSnapshot) entityKey() entityKey {
+	return entityKey{"handler", op.Snapshot.HandlerKey, op.Snapshot.InstanceID}
```
jmalloc:
```suggestion
	return entityKey{"snapshot", op.Snapshot.HandlerKey, op.Snapshot.InstanceID}
```

As above.

jmalloc (changes requested):
Hey mate, looking good!  I haven't yet delved into the individual persistence implementations, comments thus far are about the handler/persistence abstraction.

Thanks!

handler/aggregate/adaptor_test.go
```diff
@@ -248,6 +251,82 @@ var _ = Describe("type Adaptor", func() {
...
+					}, nil
+				}
+
+				adaptor.Marshaler = &codec.Marshaler{} // an empty marshaler cannot marshal anything
+
+				err := adaptor.HandleMessage(ctx, work, cause)
+				Expect(err).To(MatchError("no codecs support the '*fixtures.AggregateRoot' type"))
+			})
+
+			It("updates the snapshot if interval revision", func() {
```
jmalloc:
```suggestion
			It("updates the snapshot if the revision is a multiple of the snapshot interval", func() {
```

jmalloc:
I'm back from leave today and picking up where I left on this review. I'm def happy with the overall approach, I'm ~not~ now on to looking through the individual provider implementations.

persistence/boltpersistence/aggregate.go
```diff
@@ -55,6 +62,35 @@ func (ds *dataStore) LoadAggregateMetaData(
...
+			); ok {
+				ss := loadAggregateSnapshot(root, hk, id)
+				inst.LastEventID = ss.GetLastEventId()
+				inst.Packet.MediaType = ss.GetMediaType()
+				inst.Packet.Data = ss.GetData()
+			}
+		},
+	)
+
+	return inst, true, nil
```
jmalloc: The boolean return value here should only be `true` if an aggregate snapshot was actually found.

persistence/error.go
```diff
@@ -29,3 +31,17 @@ func (e ConflictError) Error() string {
 		e.Cause,
 	)
 }
+
+// NotFoundError is an error indicating one or more operations within a batch
+// caused a record not found error.
+type NotFoundError struct {
```
jmalloc: I don't think this error is strictly necessary:

- The `SaveAggregateSnapshot` operation should always succeed, regardless of whether there is an existing snapshot or not.
- The `RemoveAggregateSnapshot` operation does not need to produce an error if there is no existing snapshot; it is sufficient to treat this as an "idempotent assertion" that there is no snapshot for the specified instance.

persistence/internal/providertest/aggregateop.go
```diff
@@ -213,5 +215,140 @@ func declareAggregateOperationTests(tc *TestContext) {
 				))
 			})
 		})
+
+		ginkgo.Describe("type persistence.SaveAggregateSnapshot", func() {
+			ginkgo.When("the instance does not exist", func() {
```
jmalloc: Just a note for improvements to terminology throughout these tests - "instance" refers to the concept of the "aggregate instance" as defined  in [dogmatiq/dogma](https://github.com/dogmatiq/dogma#aggregate), not just the generic sense of an "instance of an object", so it's more precise use language like:

```suggestion
			ginkgo.When("there is no snapshot of the instance", func() {
```

persistence/internal/providertest/aggregaterepo.go
```diff
@@ -89,5 +90,58 @@ func declareAggregateRepositoryTests(tc *TestContext) {
 				}
 			})
 		})
+
+		ginkgo.Describe("func LoadAggregateSnapshot()", func() {
+
```
jmalloc:
```suggestion
```

persistence/memorypersistence/aggregate.go
```diff
@@ -27,6 +27,24 @@ func (ds *dataStore) LoadAggregateMetaData(
...
+
+	key := instanceKey{hk, id}
+	if inst, ok := ds.db.aggregate.snapshot[key]; ok {
+		return inst, true, nil
+	}
+
+	return persistence.AggregateSnapshot{
+		HandlerKey: hk,
+		InstanceID: id,
+	}, false, nil
```
jmalloc:
```suggestion
	ss, ok := ds.db.aggregate.snapshot[key]; ok {
	return ss, ok, nil
```

This can be simplified somewhat because unlike the `AggregateMetaData` which is always "valid" even if it hadn't been persisted, the snapshot need only be valid if it's known to already exist. Suggestion also uses the `ss` var name as you'd used elsewhere.

Note also that the fail-case branch of the existing code is not yet covered by tests - mentioning that now in case it's not as obvious after applying my suggestion.

persistence/memorypersistence/aggregate.go
```diff
@@ -46,23 +64,61 @@ func (v *validator) VisitSaveAggregateMetaData(
...
+	if x, ok := v.db.aggregate.snapshot[key]; ok {
+		if inst.LastEventID == x.LastEventID {
+			return nil
+		}
+	}
+
+	return persistence.NotFoundError{
+		Cause: op,
+	}
+}
```
jmalloc: Much the same as for `SaveAggregateSnapshot`, there's no semantic reason to reject a remove operation, so no validation is required.

I note you are checking the last event ID, which I guess serves as a kind of confirmation that you are deleting "this snapshot", but the `RemoveAggregateSnapshot` operation can be thought of as "make sure there is no snapshot of this instance" at all, since we only ever need a single snapshot.

```suggestion
	return nil
}
```

persistence/memorypersistence/aggregate.go
```diff
@@ -72,3 +128,36 @@ func (db *aggregateDatabase) save(md persistence.AggregateMetaData) {
...
+// VisitRemoveAggregateSnapshot applies the changes in a "RemoveAggregateSnapshot"
+// operation to the database.
+func (c *committer) VisitRemoveAggregateSnapshot(
+	_ context.Context,
+	op persistence.RemoveAggregateSnapshot,
+) error {
+	inst := op.Snapshot
+	key := instanceKey{inst.HandlerKey, inst.InstanceID}
+
+	c.db.process.remove(key)
```
jmalloc: Something not right here, it's referring to process removal.

jmalloc:
I'm most of the way through reviewing the BoltDB and in-memory implementations. I have some suggestions for those that will likely apply across the board. I'll have review feedback tomorrow.  

Judging by the codecov.io checks, I suspect there's no tests for the `RemoveAggregateSnapshot` operation, but I haven't gotten far enough to confirm that for myself just yet. No worries if that's intentional for now.

persistence/memorypersistence/aggregate.go
```diff
@@ -46,23 +64,61 @@ func (v *validator) VisitSaveAggregateMetaData(
...
+) error {
+	ss := op.Snapshot
+
+	if ss.HandlerKey != "" && ss.InstanceID != "" {
+		return nil
+	}
+
+	return persistence.NotFoundError{
+		Cause: op,
+	}
```
jmalloc: The `validator` need only verify that it is semantically ok to apply this operation to the database, it doesn't need to perform general import validation.  Because it's always ok to save a snapshot, this can simply return `nil`.

```suggestion
	return nil
```

jmalloc (changes requested):

jmalloc:
Just letting you know I haven't forgotten about this; as we move closer to actually writing some code in my new role I'll get back to things like this.

Evildethow:
I'm going to close this.

It's been close to a year now since this was created, and I'm guessing its diverged somewhat from `main`.

